PR #1, opened 2016-12-06
Build improvements

jml:
As suggested by @tomwilkie

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -16,7 +16,5 @@ dependencies:
 
 test:
   override:
-    - make RM= lint:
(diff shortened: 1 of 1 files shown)


tomwilkie:
Looks like building the build image outside the makefile doesn't work due to the special tricks for the go get dependancies.

jml:
> Looks like building the build image outside the makefile doesn't work due to the special tricks for the go get dependancies.

This has been fixed. PTAL.

tomwilkie:
LGTM!

PR #4, opened 2016-12-08
Correct websocket detection

jml:
Part of a fix for weaveworks/monitoring#82 -- we aren't detecting websocket handshakes correctly

### DIFF
--- a/middleware/instrument.go
+++ b/middleware/instrument.go
@@ -19,7 +19,8 @@ type Instrument struct {
 	Duration *prometheus.HistogramVec
 }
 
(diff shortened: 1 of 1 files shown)


middleware/instrument.go
```diff
@@ -20,8 +20,16 @@ type Instrument struct {
 }
 
 func isWSHandshakeRequest(req *http.Request) bool {
```
tomwilkie: Can we make this public, so literally the same code is used in authfe?

tomwilkie:
One comment, otherwise LGTM

PR #5, opened 2016-12-12
Add redirect middleware

tomwilkie:
As we're getting rid of the nginx frontend, the authfe need to redirect from scope.weave.works to cloud.w.w and http://cloud.w.w to https.

### DIFF
--- a/middleware/redirect.go
+++ b/middleware/redirect.go
@@ -15,7 +15,7 @@ type Redirect struct {
 }
 
 // Match specifies a match for a redirect.  Host and/or Scheme can be empty
-// signifiy match-all.
+// signify match-all.
 type Match struct {
 	Host, Scheme string
 }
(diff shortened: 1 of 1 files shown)


middleware/redirect.go
```diff
@@ -0,0 +1,48 @@
...
+// Matches to RedirectScheme://RedirectHost
+type Redirect struct {
+	Matches []Match
+
+	RedirectHost   string
+	RedirectScheme string
+}
+
+// Match specifies a match for a redirect.  Host and/or Scheme can be empty
+// signifiy match-all.
```
ekimekim: nit: `to signify`

ekimekim: It strikes me that `Host` can be blank. This api leaves no ability to match on blank hosts only. Or HTTP 1.0 calls without a Host header for that matter (which could probably be considered the same).
I don't see a nice way to resolve this though, so unless you feel strongly about it I'm prepared to say YAGNI.

PR #7, opened 2016-12-12
Option for logging request headers

jml:
To help us figure out what's up with weaveworks/monitoring#82

### DIFF
--- a/middleware/logging.go
+++ b/middleware/logging.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/http/httputil"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


middleware/logging.go
```diff
@@ -25,13 +27,22 @@ func (l Log) Wrap(next http.Handler) http.Handler {
 		if l.LogSuccess || !(100 <= i.statusCode && i.statusCode < 400) {
 			log.Infof("%s %s (%d) %s", r.Method, uri, i.statusCode, time.Since(begin))
 		}
+		if l.LogRequestHeaders {
+			headers, err := httputil.DumpRequest(r, false)
```
tomwilkie: May want to consider doing this before the call to `next.ServeHTTP`, as some handlers modify the request.

tomwilkie:
1 nit, LGTM otherwise

jml:
Good point. Fixed.

Issue #11, opened 2017-01-25
We depend on too much stuff


tomwilkie:
```
$ gvt fetch github.com/weaveworks/common
2017/01/25 14:21:24 Fetching: github.com/weaveworks/common
2017/01/25 14:21:26 · Skipping (existing): golang.org/x/net/context
2017/01/25 14:21:26 · Fetching recursive dependency: golang.org/x/tools/cover
2017/01/25 14:21:27 · Skipping (existing): github.com/opentracing/opentracing-go/ext
2017/01/25 14:21:27 · Skipping (existing): github.com/davecgh/go-spew/spew
2017/01/25 14:21:27 · Fetching recursive dependency: github.com/weaveworks/docker/pkg/mflag
2017/01/25 14:21:35 ·· Fetching recursive dependency: github.com/docker/docker/pkg/homedir
2017/01/25 14:22:04 ··· Fetching recursive dependency: github.com/docker/docker/vendor/github.com/opencontainers/runc/libcontainer/user
2017/01/25 14:22:04 ··· Fetching recursive dependency: github.com/docker/docker/pkg/idtools
2017/01/25 14:22:04 ···· Fetching recursive dependency: github.com/docker/docker/pkg/system
2017/01/25 14:22:04 ····· Fetching recursive dependency: github.com/docker/docker/vendor/github.com/Microsoft/go-winio
2017/01/25 14:22:04 ······ Fetching recursive dependency: github.com/docker/docker/vendor/golang.org/x/sys/windows
2017/01/25 14:22:04 ····· Fetching recursive dependency: github.com/docker/docker/vendor/github.com/docker/go-units
2017/01/25 14:22:04 ····· Fetching recursive dependency: github.com/docker/docker/vendor/github.com/Sirupsen/logrus
2017/01/25 14:22:04 ······ Fetching recursive dependency: github.com/docker/docker/vendor/golang.org/x/sys/unix
2017/01/25 14:22:05 ·· Fetching recursive dependency: github.com/docker/docker/pkg/mflag
2017/01/25 14:22:05 command "fetch" failed: error fetching github.com/docker/docker/pkg/mflag: lstat /var/folders/_b/ktq_dxhx0nbbw7gjtdzb3tn40000gn/T/gvt-769657176/pkg/mflag: no such file or directory
```

In particular, the mflag package is painful.

tomwilkie:
```
[...]

PR #16, opened 2017-01-30
Move mflag pkg here to make it easier to vendor tools.

tomwilkie:
Also, update tools from here https://github.com/weaveworks/build-tools/pull/69


tomwilkie:
Fixes #11

PR #21, opened 2017-02-28
Make the standard server shutdown gracefully.

tomwilkie:


juliusv:
:+1: though I wonder if maybe `SIGUSR1` would be better semantically than `SIGQUIT` for dumping the current goroutine traces.

PR #23, opened 2017-03-03
Instead of LogSuccess, use log levels for success vs failed requests

paulbellamy:
@tomwilkie, wdyt? as discussed the other day.

### DIFF
--- a/middleware/logging.go
+++ b/middleware/logging.go
@@ -13,7 +13,6 @@ import (
 
 // Log middleware logs http requests
 type Log struct {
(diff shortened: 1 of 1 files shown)


tomwilkie:
Looks good - could use Error instead of Warn, but up to you.

paulbellamy:
I picked warning, as it gives a bit more granularity for logging. i.e. at Error you won't get any http output at all, at Warn you'll get only failures, and at Debug you'll get everything. I was debating doing successes as Info, but we don't really enable it much on services, so...

tomwilkie:
>  I don't have a strong preference.

Me neither.  This is an improvement over what we have, so get it in.

PR #27, opened 2017-03-21
test: Do not block when reading stderr pipe

2opremio:
Needed by https://github.com/weaveworks/scope/pull/2335

I don't fully understand why it was introduced in the first place (there was no PR https://github.com/weaveworks/common/commit/201194aa4a9737826a1bdb9c3c2c38d244c0e5b3 ) but the Scope tests are happy without it.

### DIFF
--- a/test/exec/exec.go
+++ b/test/exec/exec.go
@@ -10,11 +10,6 @@ import (
 
 type mockCmd struct {
 	io.ReadCloser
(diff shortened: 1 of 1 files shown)


tomwilkie:
I think Scope is the only thing that uses this, right?

LGTM

2opremio:
I think so, yes

PR #28, opened 2017-03-31
Generalize the weaveworks/common-build image

paulbellamy:
And push it to dockerhub so that our other projects can pull it instead of building their own.

Versions are tagged from their base image, so this will push:
`weaveworks/common-build:golang-1.8.0-stretch`

If projects want to add stuff they can do a:
```Dockerfile
FROM weaveworks/common-build:golang-1.8.0-stretch
RUN whatever they need to add here
```
But, they will at least have access to all the common error checking and protobuf stuff we use without maintaining their own build image.

@tomwilkie, @jml  wdyt?


jml:
I guess I don't quite understand how it will be used. Perhaps a README update is in order?

paulbellamy:
> Readme update

Yes, definitely.

tl;dr is instead of having a `service-build`, `loki-build`, `gfdatasource-build` (maybe? this is python), you could just have one build-image with all the common stuff, that they pull instead. that way we don't need to build a separate image for each project, so building stuff should be quicker. The common image is tagged with the base image (with the go version), so that different projects could use different Go versions while upgrading.

re: circle fails, bah upgrading tools at the same time was a bad idea. will back that out.

tomwilkie:
Also, need to include common git sha in the docker tag please.

paulbellamy:
What about a tag of each? one of the base image, and the `master-f09asd0` style.

tomwilkie:
> What about a tag of each? one of the base image, and the master-f09asd0 style.

Does that work?  Can you have multiple tags?  How do you pull then?

paulbellamy:
You tag the image multiple times, and specify the tag when you're pulling

tomwilkie:
ah I see, so you would tag them `golang-1.8.0-stretch` and `golang-1.8.0-stretch-master-abcdef`?

paulbellamy:
Was thinking just `golang-1.8.0-stretch` and `master-abcdef`

paulbellamy:
I've not quite worked through how multiple different common-build images would work. (e.g. for python) maybe need to have a think about that first.

paulbellamy:
Closed in favour of: https://github.com/weaveworks/build-tools/pull/87

PR #30, opened 2017-04-21
fs: add ReadDirCount

alban:
ReadDirNames is expensive and better avoided when we don't care about
the names.

-----

This is for https://github.com/weaveworks/scope/pull/2456

TODO:
- [x] check the status of CircleCI in https://github.com/weaveworks/scope/pull/2456

### DIFF
--- a/fs/readdircount_linux_amd64.go
+++ b/fs/readdircount_linux_amd64.go
@@ -1,0 +1,62 @@
+// +build linux,amd64
+
+package fs
+
+import (
+	"fmt"
+	"os"
+	"unsafe"
+
+	"syscall"
+)
+
+func countDirEntries(buf []byte, n int) int {
+	count := 0
+	buf = buf[:n]
+	for len(buf) > 0 {
+		// see man page getdents(2) for struct linux_dirent64
+		reclenOffset := unsafe.Offsetof(syscall.Dirent{}.Reclen)
+		reclen := *(*uint16)(unsafe.Pointer(&buf[reclenOffset]))
+
+		inoOffset := unsafe.Offsetof(syscall.Dirent{}.Ino)
+		ino := *(*uint64)(unsafe.Pointer(&buf[inoOffset]))
+
+		if int(reclen) > len(buf) {
+			return count
+		}
+		buf = buf[reclen:]
+		if ino == 0 {
+			continue
+		}
+		count++
+	}
+	return count
+}
+
+// ReadDirCount is similar to ReadDirNames() and then counting with len() but
+// it is optimized to avoid parsing the entries
+func (realFS) ReadDirCount(dir string) (int, error) {
+	buf := make([]byte, 4096)
+	fh, err := os.Open(dir)
+	if err != nil {
+		return 0, err
+	}
+	defer fh.Close()
+
+	openFilesCount := 0
+	for {
+		n, err := syscall.ReadDirent(int(fh.Fd()), buf)
+		if err != nil {
+			return 0, fmt.Errorf("ReadDirent() failed: %v", err)
+		}
+		if n == 0 {
+			break
+		}
+
+		openFilesCount += countDirEntries(buf, n)
+	}
+
+	// "." and ".." don't count as files to be counted
+	nDotFiles := 2
+	return openFilesCount - nDotFiles, err
+}
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -11,6 +11,7 @@ import (
 type Interface interface {
 	ReadDir(string) ([]os.FileInfo, error)
 	ReadDirNames(string) ([]string, error)
+	ReadDirCount(string) (int, error)
 	ReadFile(string) ([]byte, error)
 	Lstat(string, *syscall.Stat_t) error
 	Stat(string, *syscall.Stat_t) error
@@ -61,6 +62,11 @@ func ReadDir(path string) ([]os.FileInfo, error) {
 // ReadDirNames see os.File.ReadDirNames
 func ReadDirNames(path string) ([]string, error) {
 	return fs.ReadDirNames(path)
+}
+
+// ReadDirCount is an optimized way to call len(ReadDirNames)
+func ReadDirCount(path string) (int, error) {
+	return fs.ReadDirCount(path)
 }
 
 // ReadFile see ioutil.ReadFile
--- a/fs/fs_test.go
+++ b/fs/fs_test.go
@@ -1,0 +1,81 @@
+package fs
+
+import (
+	"os"
+	"testing"
+)
+
+const devNullCount = 500
+
+func openDevNullFiles(b *testing.B, n int) []*os.File {
+	var arr []*os.File
+	for i := 0; i < n; i++ {
+		fh, err := os.Open("/dev/null")
+		if err != nil {
+			b.Fatalf("Cannot open /dev/null.")
+		}
+		arr = append(arr, fh)
+	}
+	return arr
+
+}
+func closeDevNullFiles(b *testing.B, arr []*os.File) {
+	for i, _ := range arr {
+		err := arr[i].Close()
+		if err != nil {
+			b.Fatalf("Cannot close /dev/null.")
+		}
+	}
+}
+
+func BenchmarkReadDirNames(b *testing.B) {
+	arr := openDevNullFiles(b, devNullCount)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		names, err := ReadDirNames("/proc/self/fd")
+		if err != nil {
+			b.Fatalf("ReadDirNames failed: %v", err)
+		}
+		count := len(names)
+		if count < devNullCount || count > devNullCount+10 {
+			b.Fatalf("ReadDirNames failed: count=%d", count)
+		}
+	}
+	b.StopTimer()
+
+	closeDevNullFiles(b, arr)
+}
+
+func BenchmarkReadDirCount(b *testing.B) {
+	arr := openDevNullFiles(b, devNullCount)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		count, err := ReadDirCount("/proc/self/fd")
+		if err != nil {
+			b.Fatalf("ReadDirCount failed: %v", err)
+		}
+		if count < devNullCount || count > devNullCount+10 {
+			b.Fatalf("ReadDirCount failed: count=%d", count)
+		}
+	}
+	b.StopTimer()
+
+	closeDevNullFiles(b, arr)
+}
+
+func TestReadDirNames(t *testing.T) {
+	names, err := ReadDirNames("/proc/self/fd")
+	if err != nil {
+		t.Fatalf("ReadDirNames failed: %v", err)
+	}
+	count, err := ReadDirCount("/proc/self/fd")
+	if err != nil {
+		t.Fatalf("ReadDirCount failed: %v", err)
+	}
+	if len(names) != count {
+		t.Fatalf("ReadDirNames and ReadDirCount give inconsitent results: %d != %d", len(names), count)
+	}
+
+}
--- a/fs/readdircount_unsupported.go
+++ b/fs/readdircount_unsupported.go
@@ -1,0 +1,9 @@
+// +build !linux !amd64
+
+package fs
+
+// ReadDirCount, unoptimized version
+func (realFS) ReadDirCount(path string) (int, error) {
+	names, err := ReadDirNames(path)
+	return len(names), err
+}
--- a/test/fs/fs.go
+++ b/test/fs/fs.go
@@ -112,6 +112,11 @@ func (p dir) ReadDirNames(path string) ([]string, error) {
 	return fs.ReadDirNames(tail)
 }
 
+func (p dir) ReadDirCount(path string) (int, error) {
+	names, err := p.ReadDirNames(path)
+	return len(names), err
+}
+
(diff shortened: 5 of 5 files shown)


alban:
@scanf I started a number of processes opening a lot of file descriptors with [openfiles.go](https://gist.github.com/alban/009763ef2794f5733aa5463e94f8cdba). And then, I used a custom program [readdirnames.go](https://gist.github.com/alban/5a6fedd06432c25e253e69f6ffe2c170) to measure the time difference between the two methods. See https://github.com/weaveworks/scope/issues/1011#issuecomment-295838863

I also looked at the cpu usage of Scope in "top", see https://github.com/weaveworks/scope/pull/2456

scanf:
@alban Thanks for the detailed explanation.

fs/readdircount_linux_amd64.go
```diff
@@ -0,0 +1,67 @@
...
+		buf = buf[reclen:]
+		if ino == 0 {
+			continue
+		}
+		count++
+	}
+	return count
+}
+
+func getdents64(fd int, buf []byte) (n int, err int) {
```
2opremio: Please use https://godoc.org/golang.org/x/sys/unix#Getdents instead

fs/readdircount_linux_amd64.go
```diff
@@ -0,0 +1,67 @@
...
+	"syscall"
+	"unsafe"
+)
+
+func countDirEntries(buf []byte, n int) int {
+	count := 0
+	buf = buf[:n]
+	for len(buf) > 0 {
+		// see man page getdents(2) for struct linux_dirent64
+		reclen := *(*uint16)(unsafe.Pointer(&buf[16]))
```
2opremio: Please use [`unsafe.Offsetof(Dirent{}.Reclen)`](https://github.com/golang/sys/blob/493114f68206f85e7e333beccfabc11e98cba8dd/unix/syscall_linux.go#L1002)  instead

IN fact,

fs/readdircount_linux_amd64.go
```diff
@@ -0,0 +1,67 @@
...
+	r0, _, errno := syscall.Syscall(syscall.SYS_GETDENTS64, uintptr(fd), uintptr(_p0), uintptr(len(buf)))
+	n = int(r0)
+	err = int(errno)
+	return
+}
+
+// ReadDirCount is similar to ReadDirNames() and then counting with len() but
+// it is optimized to avoid parsing the entries
+func (realFS) ReadDirCount(dir string) (int, error) {
+	buf := make([]byte, 4096)
```
2opremio: Have you played with this buffer size? [Golang defaults it to 4096](https://github.com/golang/go/blob/c05b06a12d005f50e4776095a60d6bd9c2c91fac/src/os/dir_unix.go#L16) but, while we are it, Scope may benefit from an even larger size (I haven't done the math though)

fs/readdircount_linux_amd64.go
```diff
@@ -0,0 +1,67 @@
...
+	"unsafe"
+)
+
+func countDirEntries(buf []byte, n int) int {
+	count := 0
+	buf = buf[:n]
+	for len(buf) > 0 {
+		// see man page getdents(2) for struct linux_dirent64
+		reclen := *(*uint16)(unsafe.Pointer(&buf[16]))
+		ino := *(*uint64)(unsafe.Pointer(&buf[0]))
```
2opremio: Same  here

fs/readdircount_linux_amd64.go
```diff
@@ -0,0 +1,67 @@
...
+
+		if int(reclen) > len(buf) {
+			return count
+		}
+		buf = buf[reclen:]
+		if ino == 0 {
```
2opremio: I guess this is the inode of `.` and `..`? Please add a comment

fs/readdircount_linux_amd64.go
2opremio: Actually I think you can use https://golang.org/pkg/syscall/#Getdents directly

2opremio: Or even https://golang.org/pkg/syscall/#ReadDirent

fs/readdircount_linux_amd64.go
2opremio: Also, we may benefit from reusing the buffer (buffer pool?)

alban: I have not compared the perfs with different sizes.

On my laptop, Chrome has ~500 opened files and that was the only process where 4096 was not big enough to read everything in one go. My code had a bug initially in that case. I tried a small buffer (100) to reproduce the issue and to be able to check if the loop works correctly in that case.

fs/readdircount_linux_amd64.go
alban: I am not using `syscall.Dirent` (== `struct linux_dirent` in C) but `struct linux_dirent64`. Both are described in man page getdents(2). But the latter does not have a Go struct ready to use, unfortunately.

But fortunately for me, the first two fields are 64 bits, so there should be no alignment problems:
```
           struct linux_dirent64 {
               ino64_t        d_ino;    /* 64-bit inode number */
               off64_t        d_off;    /* 64-bit offset to next structure */
               unsigned short d_reclen; /* Size of this dirent */
               unsigned char  d_type;   /* File type */
               char           d_name[]; /* Filename (null-terminated) */
           };
```

fs/readdircount_linux_amd64.go
alban: No, `.` and `..` should have real inodes numbers. AFAIU the case `ino == 0` could happen when there is no files on that directory entry (there is a hole in the record). I don't think that procfs returns such entries, but `ReadDirCount()` is supposed to be generic. I took that from:
https://golang.org/src/syscall/dirent.go?s=2134:2232#L78
```
    78		if ino == 0 { // File absent in directory.
    79			continue
    80		}
    81	
```

fs/readdircount_linux_amd64.go
2opremio: Then I think it's worth playing with bigger sizes.

2opremio: (i.e. Think about server processes with a lot of open files)

fs/readdircount_linux_amd64.go
2opremio: OK, please add a comment.

> I don't think that procfs returns such entries

It seems it does, so please adjust the code accordingly.

```
$ ls -la /proc/self/fd/
total 0
dr-x------ 2 ubuntu ubuntu  0 Apr 25 14:18 .
dr-xr-xr-x 9 ubuntu ubuntu  0 Apr 25 14:18 ..
lrwx------ 1 ubuntu ubuntu 64 Apr 25 14:18 0 -> /dev/pts/1
lrwx------ 1 ubuntu ubuntu 64 Apr 25 14:18 1 -> /dev/pts/1
lrwx------ 1 ubuntu ubuntu 64 Apr 25 14:18 2 -> /dev/pts/1
lr-x------ 1 ubuntu ubuntu 64 Apr 25 14:18 3 -> /proc/12025/fd
```

How did you test this? I am surprised you didn't notice the off-by-2 in the results.

alban: I meant I don't think procfs returns entries with "ino == 0". It returns the "." and ".." but those have valid inode numbers.

```
$ ls -lai /proc/self/fd/
total 0
5470776 dr-x------. 2 alban alban  0 Apr 25 16:22 .
5470775 dr-xr-xr-x. 9 alban alban  0 Apr 25 16:22 ..
5470781 lrwx------. 1 alban alban 64 Apr 25 16:22 0 -> /dev/pts/1
5470782 lrwx------. 1 alban alban 64 Apr 25 16:22 1 -> /dev/pts/1
5470783 lrwx------. 1 alban alban 64 Apr 25 16:22 2 -> /dev/pts/1
5470784 lr-x------. 1 alban alban 64 Apr 25 16:22 3 -> /proc/29189/fd
```
The off-by-2 is done in `ReadDirCount()` with `nDotFiles := 2`. I tested manually by running Scope and looking at a few processes with `ls -l /proc/<pid>/fd/ | wc -l`.

2opremio: Man, I'm thick today. My apologies and thanks for the clarifications.

fs/readdircount_linux_amd64.go
2opremio: In amd64, `syscall.Dirent` maps to `struct linux_dirent64`, see https://golang.org/src/syscall/ztypes_linux_amd64.go?s=1903:2035#L124

2opremio: Otherwise `syscall.Getdents` wouldn't work

`Getdents` in amd64 is the 64bit version (see https://github.com/golang/go/blob/5a303aa1e96b669051a18bdc44465396a45037ea/src/syscall/syscall_linux.go#L759 , https://golang.org/src/syscall/zsyscall_linux_amd64.go?s=12082:12134#L491 and https://github.com/golang/go/blob/9e6b79a5dfb2f6fe4301ced956419a0da83bd025/src/syscall/syscall_linux_amd64.go#L9 ) and the parsing of dirents relays on the offsets of `syscall.Dirent{}` (see https://github.com/golang/go/blob/9ed0715bb66dbbd0f597f93d0bc70a3d769b1b10/src/syscall/dirent.go#L68 and https://github.com/golang/go/blob/5a303aa1e96b669051a18bdc44465396a45037ea/src/syscall/syscall_linux.go#L768 )

fs/readdircount_linux_amd64.go
alban: Thanks, ReadDirent works.

fs/readdircount_linux_amd64.go
alban: updated

alban:
Rebased and updated. I added a commit with a small test and benchmarks:
```
$ go test -bench Benchmark*
BenchmarkReadDirNames-4            10000            155566 ns/op
BenchmarkReadDirCount-4            10000            124618 ns/op
```
So we can check that ReadDirCount is faster than ReadDirNames without using Scope or another external program.

fs/readdircount_linux_amd64.go
```diff
@@ -0,0 +1,62 @@
...
+func countDirEntries(buf []byte, n int) int {
+	count := 0
+	buf = buf[:n]
+	for len(buf) > 0 {
+		// see man page getdents(2) for struct linux_dirent64
+		reclenOffset := unsafe.Offsetof(syscall.Dirent{}.Reclen)
+		reclen := *(*uint16)(unsafe.Pointer(&buf[reclenOffset]))
+
+		inoOffset := unsafe.Offsetof(syscall.Dirent{}.Ino)
+		ino := *(*uint64)(unsafe.Pointer(&buf[inoOffset]))
```
2opremio: There is no need to use unsafe pointers , see https://github.com/golang/go/blob/9ed0715bb66dbbd0f597f93d0bc70a3d769b1b10/src/syscall/dirent.go#L50

I wonder if we also need the bounds hint

fs/readdircount_linux_amd64.go
```diff
@@ -0,0 +1,62 @@
...
+	"syscall"
+)
+
+func countDirEntries(buf []byte, n int) int {
+	count := 0
+	buf = buf[:n]
+	for len(buf) > 0 {
+		// see man page getdents(2) for struct linux_dirent64
+		reclenOffset := unsafe.Offsetof(syscall.Dirent{}.Reclen)
+		reclen := *(*uint16)(unsafe.Pointer(&buf[reclenOffset]))
```
2opremio: Same here

fs/readdircount_linux_amd64.go
2opremio: Did you get to play with this?

fs/readdircount_linux_amd64.go
alban: Without unsafe.Pointer, I get:
```
./readdircount_linux_amd64.go:19: cannot convert &buf[reclenOffset] (type *byte) to type *uint16
```
I think the situation is different than in your link because `uint64(b[0])` is not reading the pointer but the value at index 0.

We need to specify the uint16 / uint64  so that it takes the correct amount of bytes from the array to form the integer. Otherwise, it would just take one byte because `buf` is a `[]byte`

fs/readdircount_linux_amd64.go
alban: In a buffer of 4096 bytes, we can fit 170 entries (it's linear).

I tried the benchmarks with 2048 entries (increasing `ulimit -n` and opening `/dev/null` 2048 times to have that many entries in `/proc/self/fd`):
- len(buf)=64   ->  972097 ns/op
- len(buf)=256   ->  694169 ns/op
- len(buf)=512   ->  578525 ns/op
- len(buf)=1024 ->  570824 ns/op
- len(buf)=4096 ->  597399 ns/op
- len(buf)=8192 ->  573534 ns/op

Conclusion: that varies a bit but when the buffer is too small, the performance is negatively impacted. The buffer needs to be at least at 512 bytes (fitting ~21 directory entries), but increasing its size further does not improve the performance further much. I suggest to keep it at 4096.

I have not tried buffer pools yet.

fs/readdircount_linux_amd64.go
2opremio: Fair enough

fs/readdircount_linux_amd64.go
2opremio: Probably not worth the effort for now

PR #31, opened 2017-04-25
Update HTTPRequest extract functions to make it clear which ID they are extracting

aaron7:
There are plans to rename "orgID" in weave cloud to "instanceID" and "userID" in cortex to "principalID".

This change makes it clearer which ID these functions are extracting in the meantime.


jml:
You probably need to rebase this on master to get the test fixes.

PR #33, opened 2017-04-27
Use orgID and userID for the user context

aaron7:
We have been using the term `userID` to refer to instances (which we call `orgID` elsewhere in our codebase, e.g. headers). We wish to use both `orgID` and `userID` (the individual) in a new service.

This change renames `userID` to `orgID` and creates a new context key for actual `userID`'s (the individuals).

I have proxied the old functions to the `orgID` ones to not break dependant code.

### DIFF
--- a/user/http.go
+++ b/user/http.go
@@ -16,7 +16,7 @@ const (
 )
 
 // ExtractOrgIDFromHTTPRequest extracts the org ID from the request headers and returns
-// the org ID and a context with the org ID embbedded.
+// the org ID and a context with the org ID embedded.
 func ExtractOrgIDFromHTTPRequest(r *http.Request) (string, context.Context, error) {
 	orgID := r.Header.Get(orgIDHeaderName)
 	if orgID == "" {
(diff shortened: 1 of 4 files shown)


user/http.go
```diff
@@ -3,37 +3,30 @@ package user
...
 
-// Errors that we return
-const (
-	ErrNoOrgID               = errors.Error("no org id")
-	ErrDifferentOrgIDPresent = errors.Error("different org ID already present")
-)
-
-// ExtractFromHTTPRequest extracts the org ID from the request headers and returns
+// ExtractOrgIDFromHTTPRequest extracts the org ID from the request headers and returns
 // the org ID and a context with the org ID embbedded.
```
jml: "embedded"

user/http.go
```diff
@@ -45,3 +38,37 @@ func InjectIntoHTTPRequest(ctx context.Context, r *http.Request) error {
 	r.Header.Set(orgIDHeaderName, orgID)
 	return nil
 }
+
+// ExtractUserIDFromHTTPRequest extracts the org ID from the request headers and returns
+// the org ID and a context with the org ID embbedded.
```
jml: "embedded"

user/id.go
```diff
@@ -9,27 +9,56 @@ import (
...
+
+// Extract is deprecated. Use ExtractOrgID.
+func Extract(ctx context.Context) (string, error) {
+	return ExtractOrgID(ctx)
+}
+
+// Inject is deprecated. Use returns a derived context containing the org ID.
+func Inject(ctx context.Context, orgID string) context.Context {
+	return InjectOrgID(ctx, orgID)
+}
```
jml: We don't need to keep the old versions around. When we can update the call sites when we vendor this version of common.

jml:
I reckon go ahead & merge this.

PR #34, opened 2017-05-01
middleware/path_rewrite: Don't inadvertently unescape url-encoded characters

ekimekim:
When a URL is parsed, the original url-encoded path goes in URL.RawPath,
and the decoded path in URL.Path.

URL's string method will use URL.RawPath, preserving the original encoding,
but only if URL.RawPath is a valid encoding of URL.Path.

If we modify URL.Path directly, this is no longer true, and URL then uses the un-escaped version of Path.
This has adverse effects since meaning can differ - for example, if the path contains an encoded / character
(ie. "%2F"), we want to preserve this as an encoded character, as otherwise it appears to be a path seperator.
"/foo%2Fbar" is not the same path as "/foo/bar".

We work around this by applying the regex to the escaped version instead, then reconstructing the unescaped version
from the new escaped version.

The new escaped version SHOULD be a valid url-encoded string - any bad user input would have already caused a 400
during initial parsing. However, there exists a possibility that a regex could produce an invalid result from a valid
input. Since this case is a programmer error, we log.Errorf() and return a 500.

### DIFF
--- a/middleware/path_rewrite.go
+++ b/middleware/path_rewrite.go
@@ -2,7 +2,10 @@ package middleware
 
 import (
 	"net/http"
+	"net/url"
 	"regexp"
+
+	log "github.com/Sirupsen/logrus"
 )
 
 // PathRewrite supports regex matching and replace on Request URIs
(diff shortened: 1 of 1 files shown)


ekimekim:
The linter is failing on files I didn't touch?

jml:
Yep. @2opremio merged #30, which looks like it broke the build:

```
./tools/lint -notestpackage -ignorespelling queriers -ignorespelling Queriers .

./fs/fs_test.go: run gofmt -s -w ./fs/fs_test.go

./fs/fs_test.go:23:9: should omit 2nd value from range; this loop is equivalent to `for i := range ...`

./mflag/flag.go: higher than 25 cyclomatic complexity - 30 mflag (*FlagSet).parseOne ./mflag/flag.go:984:1

./mflag/flag_test.go: higher than 25 cyclomatic complexity - 27 mflag testParse ./mflag/flag_test.go:153:1
```

tomwilkie:
But otherwise LGTM!  I'll fixup the errors on master.

tomwilkie:
If you rebase, master is fixed now.

alban:
@ekimekim [Golang's url.PathUnescape](https://github.com/golang/go/commit/7e2bf952a905f16a17099970392ea17545cdd193) was added in Go 1.8 but [Scope uses Go 1.7](https://github.com/weaveworks/scope/blob/master/backend/Dockerfile#L3). Should Scope be updated to use the latest Go version?

Meanwhile I could vendor github.com/weaveworks/common in Scope without this PR (to take https://github.com/weaveworks/common/pull/30).

2opremio:
> Yep. @2opremio merged #30, which looks like it broke the build

Yep, my bad, we should enable CircleCI for external PRs (I don't think we have secrets in the current setup)

ekimekim:
@alban I see no reason we shouldn't update to 1.8, don't want to do that without a more thorough investigation though. Yeah, in the meantime update without this PR. We'll either do the upgrade or I can work around the lack of that function somehow.

PR #37, opened 2017-05-25
Make it easy to get a logger with context information

jml:
Related to https://github.com/weaveworks/service/issues/1233

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -6,7 +6,10 @@ import (
 	"os"
 	"strings"
 
+	"golang.org/x/net/context"
+
 	log "github.com/Sirupsen/logrus"
+	"github.com/weaveworks/common/user"
 )
 
 // Setup configures logging output to stderr, sets the log level and sets the formatter.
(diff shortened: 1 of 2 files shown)


logging/logging.go
```diff
@@ -43,3 +46,21 @@ func (f *textFormatter) Format(entry *log.Entry) ([]byte, error) {
 	b.WriteByte('\n')
 	return b.Bytes(), nil
 }
+
+// WithContext returns a logger that has information from the context as fields.
+//
+// e.g.
+//     logger := logging.WithContext(ctx)
+//     logger.Errorf("Some error")
+func WithContext(ctx context.Context) *log.Entry {
```
tomwilkie: `WithUsedIDFromContext` might be more appropriate?  or `WithUsedID`?

jml: I figure if ever we add more information to the context, we might want to add more information here. `WithStandardWeaveworksContextInfo` is a bit wordy.

I guess instead we could add more `WithFoo` functions, since they do chain.

tomwilkie: How do they chain?

TBH I'd prefer if these lived in their various packages - ie this was in `users` as `Logger`, and there was one in the tracing package etc.  These could then be combined in a standard weaveworks logger here.  But just a thought.

tomwilkie:
LGTM

PR #41, opened 2017-06-02
Log request headers, response headers and some body as warning on 500s

jml:
Part of https://github.com/weaveworks/service/issues/1233 

This is kind of a significant change, as it intercepts *all* incoming requests. I haven't done much in the way of testing.

Also changes the logging order slightly. Previously, when `LogRequestHeaders` was on, we would log headers then request/response summary (method, status, etc). Now, we log method, status, etc. _then_ headers.

### DIFF
--- a/middleware/logging.go
+++ b/middleware/logging.go
@@ -16,7 +16,6 @@ import (
 // Log middleware logs http requests
 type Log struct {
 	LogRequestHeaders bool // LogRequestHeaders true -> dump http headers at debug log level
-
 }
 
 // logWithRequest information from the request and context as fields.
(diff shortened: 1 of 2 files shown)


middleware/response.go
```diff
@@ -0,0 +1,157 @@
...
+	bodyBytesLeft int
+}
+
+// newBadResponseLogger makes a new badResponseLogger.
+func newBadResponseLogger() *badResponseLogger {
+	return &badResponseLogger{
+		recorder:      httptest.NewRecorder(),
+		logBody:       false,
+		bodyBytesLeft: maxResponseBodyInLogs,
+		statusCode:    http.StatusOK,
```
bboreham: Seems odd to me that status is OK if neither `Write` or `WriteHeader` are called.  I guess this ends up shortening the `Write`-with-no-`WriteHeader` case; if that is the reason maybe comment that?

middleware/response.go
```diff
@@ -0,0 +1,157 @@
...
+	// track whether or not we've called WriteHeader.
+	if !b.logBody {
+		// We don't need to log anything, so lie and say that we wrote
+		// everything, making this effectively a no-op.
+		return len(data), nil
+	}
+
+	if len(data) > b.bodyBytesLeft {
+		b.recorder.Write(data[:b.bodyBytesLeft])
+		b.recorder.WriteString("…")
```
bboreham: I don't think we want to do this more than once; maybe reset `logBody` at this point?

bboreham: Question the use of Unicode ellipsis rather than three dots; it assumes the operator has a certain level of tooling.

middleware/response.go
```diff
@@ -0,0 +1,157 @@
...
+	return httputil.DumpResponse(b.recorder.Result(), true)
+}
+
+// Header implements http.ResponseWriter.
+func (b *badResponseLogger) Header() http.Header {
+	return b.recorder.Header()
+}
+
+// WriteHeader implements http.ResponseWriter. It will immediately log the
+// response headers if `statusCode` is a 5XX.
```
bboreham: I'm not clear what this comment means.  Is it just out of date?

middleware/logging.go
```diff
@@ -9,33 +9,48 @@ import (
...
+			if l.LogRequestHeaders && headers != nil {
+				logWithRequest(r).Debugf("Is websocket request: %v\n%s", IsWSHandshakeRequest(r), string(headers))
+			}
 		} else {
-			log.Warnf("%s %s (%d) %s", r.Method, uri, i.statusCode, time.Since(begin))
+			logWithRequest(r).Warnf("%s %s (%d) %s", r.Method, uri, i.statusCode, time.Since(begin))
+			logWithRequest(r).Warnf("Is websocket request: %v\n%s", IsWSHandshakeRequest(r), string(headers))
+			response, err := i.dumpResponse()
+			if err != nil {
+				logWithRequest(r).Errorf("Could not dump response headers: %v", err)
```
bboreham: we're not just dumping headers

middleware/response.go
jml: I have no idea what I meant by that. Have removed second sentence.

middleware/response.go
jml: Looking at the code, the `statusCode` member is unused after being set. I've removed it from the struct entirely.

jml:
>  it goes to a lot of work to capture response headers when they are not asked for in the original issue. I can't immediately think why they would be useful in a 5xx.

Does it really? It's been a while since I've looked at this, but I thought most of the work was around making sure we're still streaming headers & body, and that we don't accidentally store massive bodies in memory while building up a log entry.

I'm not sure what I'd change if we wanted to stop capturing response headers. Maybe change `badResponseLogger.recorder` to be a simple `io.Writer` rather than an `httptest.ResponseRecorder`. Or is there some other more significant simplification I'm missing?

bboreham:
> work

meaning we build a data structure of headers on _every_ response, including the 200s and 400s that we're not going to log

> change badResponseLogger.recorder to be a simple io.Writer rather than an httptest.ResponseRecorder. 

Exactly. Like a `bytes.Buffer`.

jml:
Makes sense. Doing that means that `MultiResponseWriter` is surplus to requirements and that a more specialised approach is warranted. Have merged `badResponseLogger` and `MultiResponseWriter`, which results in same behaviour with much less code.

jml:
PTAL @bboreham

bboreham:
Suggest you note this PR makes a quiet change, in that the order of request and headers is reversed in the logs, when `LogRequestHeaders` is on.

PR #45, opened 2017-07-13
Include the orgID when logging gRPC calls

bboreham:
And also emit the method and duration as fields so they can be parsed out

The code to generate the log entry is adapted from https://github.com/weaveworks/cortex/pull/452, which does the same thing for the Prometheus log interface.

### DIFF
--- a/middleware/grpc_logging.go
+++ b/middleware/grpc_logging.go
@@ -6,6 +6,8 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
+
+	"github.com/weaveworks/common/logging"
 )
 
 const gRPC = "gRPC"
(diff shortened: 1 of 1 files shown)


middleware/grpc_logging.go
```diff
@@ -14,10 +16,34 @@ const gRPC = "gRPC"
...
 }
+
+// LogEntryFromContext returns a logrus Entry that has information
+// about the current user in its details.
+//
+// e.g.
+//   entry := middleware.LogEntryFromContext(ctx)
+//   entry.Errorf("Could not chunk chunks: %v", err)
+func LogEntryFromContext(ctx context.Context) *log.Entry {
+	// Weaveworks uses "orgs" and "orgID" to represent Cortex users,
```
rade: this probably shouldn't mention cortex, since that's just one specific use.

middleware/grpc_logging.go
```diff
@@ -14,10 +16,34 @@ const gRPC = "gRPC"
...
+	}
+	return LogEntryFromUserID(userID)
+}
+
+// LogEntryFromUserID returns a log Entry that has information about
+// the current user in its details.
+func LogEntryFromUserID(userID string) *log.Entry {
+	// See note in WithContext.
+	return log.NewEntry(log.StandardLogger()).WithField("orgID", userID)
+}
```
rade: could we inline `LogEntryFromUserID` and hence dispense with all the org/user confusion and just stick to org?

bboreham: Yes we could inline it. Unless we subsequently need to log the id in a place where we don't have a Context. 

I don't think removing those comments will help the confusion much, overall.

rade: > I don't think removing those comments will help the confusion much, overall.

Removing the comments alone wouldn't. But inlining and removing them would since then the code straightforwardly extracts the org id and logs it. No mention of any user id.

leth:
Have you seen https://github.com/weaveworks/common/blob/master/logging/logging.go#L55 and https://github.com/weaveworks/common/blob/master/user/grpc.go#L25 ?

PR #47, opened 2017-07-20
Add option to not register instrumentation routes.

tomwilkie:
So Server can be used for internet-facing services.

Default of false preserves existing behaviour.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -40,6 +40,8 @@ type Config struct {
 	HTTPListenPort   int
 	GRPCListenPort   int
 
+	DontRegisterInstrumentation bool
+
 	ServerGracefulShutdownTimeout time.Duration
 	HTTPServerReadTimeout         time.Duration
 	HTTPServerWriteTimeout        time.Duration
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -40,6 +40,8 @@ type Config struct {
 	HTTPListenPort   int
 	GRPCListenPort   int
 
+	DontRegisterInstrumentation bool
```
jml: Can you please call this `RegisterInstrumentation` and invert the conditional?

tomwilkie: I did it this way so the zero value preserved the current behaviour.  This isn't something I made a flag for as I don't expect people will need to set this on the command line, and therefore I have no way of setting a default value of true.

I could make a flag with a default of true if you think that would be better - WDYT?

jml: Ah, gotcha.

I'm leaning toward flag, because one person's internet-facing service is another person's backend service. I don't feel strongly about it though. LMK if you'd rather not, and I'll hit the button.

tomwilkie: I'll make a flag, give me a sec.

jml:
Thanks!

PR #48, opened 2017-07-21
logging: Add ability to deduplicate log messages

ekimekim:
Motivating example is to not excessively log the same http 4xx code over and over from the same ip,
as this doesn't add info and bloats our log volume.

This requires some abuse of logrus to work, but we do this via a formatter which can either emit the original message,
or the empty string to log no message (this is harmless). Later, a seperate goroutine will emit a new log
that logs the original log and how many times it occurred.

We always log the first time we see a given message, then start a timer. If we see the message again at least once in that
time frame, we log at the end of the time frame how many times.
Note this means we can say "repeated once", which is a little confusing but preferable to displaying the original message
unchanged but time delayed, which can be very misleading as there's no indication of the time delay.

Fixes https://github.com/weaveworks/service-conf/issues/968

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -5,11 +5,17 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync"
+	"time"
 
 	"golang.org/x/net/context"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/weaveworks/common/user"
+)
+
+const (
+	defaultDedupeInterval = time.Minute
 )
 
 // Setup configures logging output to stderr, sets the log level and sets the formatter.
@@ -24,6 +30,39 @@ func Setup(logLevel string) error {
 	return nil
 }
 
+// SetupDeduplication should be performed after any other logging setup.
+// For all logs less severe or equal to the given log level (but still higher than the logger's configured log level),
+// these logs will be 'deduplicated'. What this means is that, excluding certain special fields like time, multiple
+// identical log entries will be grouped up and a summary message emitted.
+// For example, instead of:
+//     00:00:00 INFO User 123 did xyz
+//     00:00:10 INFO User 123 did xyz
+//     00:00:25 INFO User 123 did xyz
+//     00:00:55 INFO User 123 did xyz
+// you would get:
+//     00:00:00 INFO User 123 did xyz
+//     00:01:00 INFO Repeated 3 times: User 123 did xyz
+// The interval argument controls how long to wait for additional messages to arrive before reporting.
+// Increase it to deduplicate more aggressively, decrease it to lower latency from a log occurring to it appearing.
+// Set it to 0 to pick a sensible default value (recommended).
+// NOTE: For simplicity and efficiency, fields are considered 'equal' if and only if their string representations (%v) are equal.
+func SetupDeduplication(logLevel string, interval time.Duration) error {
+	dedupeLevel, err := log.ParseLevel(logLevel)
+	if err != nil {
+		return fmt.Errorf("Error parsing log level: %v", err)
+	}
+	if interval <= 0 {
+		interval = defaultDedupeInterval
+	}
+
+	// We use a special Formatter to either format the log using the original formatter, or to return ""
+	// so nothing will be written for that event. The repeated entries are later logged along with a field flag
+	// that tells the formatter to ignore the message.
+	stdLogger := log.StandardLogger()
+	stdLogger.Formatter = newDedupeFormatter(stdLogger.Formatter, dedupeLevel, interval)
+	return nil
+}
+
 type textFormatter struct{}
 
 // Based off logrus.TextFormatter, which behaves completely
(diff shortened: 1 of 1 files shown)


ekimekim:
Fixes https://github.com/weaveworks/service-conf/issues/968

logging/logging.go
```diff
@@ -5,13 +5,19 @@ import (
...
+	"time"
 
 	"golang.org/x/net/context"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/weaveworks/common/user"
 )
 
+const (
+	DEFAULT_DEDUPE_INTERVAL = time.Minute
```
jml: ./logging/logging.go:18:2: don't use ALL_CAPS in Go names; use CamelCase
./logging/logging.go:18:2: exported const DEFAULT_DEDUPE_INTERVAL should have comment (or a comment on this block) or be unexported

logging/logging.go
```diff
@@ -55,3 +92,94 @@ func (f *textFormatter) Format(entry *log.Entry) ([]byte, error) {
...
+		return true
+	}
+	key := fmt.Sprintf("%s %s", entry.Message, fieldsToString(entry.Data))
+	f.lock.Lock()
+	defer f.lock.Unlock()
+	if ec, ok := f.seen[key]; ok {
+		// already seen, increment count and do not log
+		ec.count++
+		f.seen[key] = ec
+		return false
```
jml: ./logging/logging.go:143:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)

logging/logging.go
```diff
@@ -24,6 +30,39 @@ func Setup(logLevel string) error {
...
+//     00:00:00 INFO User 123 did xyz
+//     00:00:10 INFO User 123 did xyz
+//     00:00:25 INFO User 123 did xyz
+//     00:00:55 INFO User 123 did xyz
+// you would get:
+//     00:00:00 INFO User 123 did xyz
+//     00:01:00 INFO Repeated 3 times: User 123 did xyz
+// The interval argument controls how many seconds to wait for additional messages to arrive before reporting.
+// Increase it to deduplicate more aggressively, decrease it to lower latency from a log occurring to it appearing.
+// Set it to 0 to pick a sensible default value (reccomended).
```
jml: ./logging/logging.go:47:49: "reccomended" is a misspelling of "recommended"

logging/logging.go
```diff
@@ -24,6 +30,39 @@ func Setup(logLevel string) error {
...
+// identical log entries will be grouped up and a summary message emitted.
+// For example, instead of:
+//     00:00:00 INFO User 123 did xyz
+//     00:00:10 INFO User 123 did xyz
+//     00:00:25 INFO User 123 did xyz
+//     00:00:55 INFO User 123 did xyz
+// you would get:
+//     00:00:00 INFO User 123 did xyz
+//     00:01:00 INFO Repeated 3 times: User 123 did xyz
+// The interval argument controls how many seconds to wait for additional messages to arrive before reporting.
```
jml: s/many seconds/long/

Because it's a `Duration` type, exact time unit doesn't matter.

logging/logging.go
```diff
@@ -24,6 +30,39 @@ func Setup(logLevel string) error {
...
+		return fmt.Errorf("Error parsing log level: %v", err)
+	}
+	if interval <= 0 {
+		interval = DEFAULT_DEDUPE_INTERVAL
+	}
+
+	// We use a special Formatter to either format the log using the original formatter, or to return ""
+	// so nothing will be written for that event. The repeated entries are later logged along with a field flag
+	// that tells the formatter to ignore the message.
+	stdLogger := log.StandardLogger()
```
jml: How does this work? If you're assigning with `:=`, then aren't you just creating a local variable called `stdLogger`? How does this propagate to the rest of the system?

ekimekim: `log.StandardLogger` returns a `*Logger`, ie. a pointer.

jml: Ahhh, gotcha.

leth:
IIRC the motivator for this was the 429 logs coming from cortex.
We've since [added the series name to the log](https://github.com/weaveworks/cortex/pull/515)
Would each different series log message be throttled separately?
Would that mean that this doesn't help with the 429s?

ekimekim:
Yes, this would mean each series is still logged, though if the same series were logged more than once per minute it would still help. We could:
* Increase the dedup interval in authfe so that the same series won't be logged again for, say, 10 minutes instead of 1. That's probably a larger latency than I'm comfortable with, however.
* Add some functionality to specify 'fields to ignore' in the dedup, though that massively complicates things and could result in surprising behaviour, and we'd still need to refactor the logs in question to only include the series in the field data and NOT the message, since we really do need to discriminate on the message text.

ekimekim:
@jml PTAL, and what are your thoughts on https://github.com/weaveworks/common/pull/48#issuecomment-318085020?

jml:
> what are your thoughts on #48 (comment)?

First, that merging this PR as-is is an incremental improvement so we should just do it.

I think @ekimekim's analysis is spot on.

The recent change adds both the series name (e.g. `http_request_duration_seconds`) and the full metric label set (e.g. `http_request_duration_seconds{path="/foo/bar",method="get"}`) in the log. The latter was extremely useful in diagnosing the client-side issue and communicating that to the customer, but I only really needed a couple of examples.

I'd be happy to refactor the logging code in Cortex to work with whatever we decide.

I wonder if we can get away with something simpler than a 'fields to ignore' that massively complicates things, e.g.

* a whitelist of fields to respect, rather than a blacklist of things to ignore. Sure, this will cause us to miss some things, but we can iterate as we discover them.
* a convention such that fields beginning with `_` get ignored (I like this less, because it's harder to explain for open source projects like Cortex, but maybe it could be a thing)

jml:
Does that help at all?

ekimekim:
> The latter was extremely useful in diagnosing the client-side issue and communicating that to the customer, but I only really needed a couple of examples.

Sounds like a good use case for a DEBUG level log instead of an INFO, surely?

As an aside, do we have the capability to, say, log at DEBUG at the service but only forward INFO and above to loggly? Then we could still get debug logs with kubectl if needed.

Anyway, merging this.

PR #60, opened 2017-09-20
Add LogContext middleware to inject org/user id

rndstr:
We want to inject the organizationID and userID from named url path parameters in multiple places which this middleware helps with.

Not sure about the naming... open to alternatives.


leth:
In other places we've sent the UserID and OrgID as HTTP headers, any reason for the departure?

rndstr:
@leth good point, I've seen both. Who is adding these HTTP headers? Clients? Authfe? And is there a middleware to extract it into the request context?

leth:
Fairly sure it's authfe; though the middleware lives in the users directory: https://github.com/weaveworks/service/blob/master/users/client/middleware.go

jml:
Sometimes in other clients directly too: https://github.com/weaveworks/cortex/blob/master/pkg/ruler/ruler.go#L223

middleware/logging.go
```diff
@@ -85,3 +86,28 @@ func (i *interceptor) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 	}
 	return hj.Hijack()
 }
+
+// LogContext is a middleware to inject the organization and user id into
```
jml: FWIW, this doesn't really have anything to do with logging. `logging.With` is expected to take a general context and add whatever useful bits it can find in the context to the logs.

This is essentially https://github.com/weaveworks/common/blob/master/middleware/http_auth.go

rndstr:
Main misconception was that there was no injection (since I didn't run it through authfe locally). Another revelation was that `orgID != externalOrgID`, so I'm closing this. Will also revert weaveworks/billing#180 and create a PR with minor improvements I stumbled upon.

PR #62, opened 2017-09-22
Add notification client library

jpellizzari:
Fixes https://github.com/weaveworks/notification/issues/21

Adds a first iteration of a client library for notifications. This intentionally contains very little abstraction and only be used for an initial non-customer-facing service, such as `service-ui-kicker`.

### DIFF
--- a/notifications/README.md
+++ b/notifications/README.md
@@ -1,0 +1,49 @@
+## Notifications
+
+The `notifications` package is a client library for the [weaveworks/notifications](https://github.com/weaveworks/notification) service.
+
+
+```golang
+package main
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/weaveworks/common/notifications"
+)
+
+func main() {
+	url := "http://eventmanager.notification.svc.cluster.local.:80"
+	sender := notifications.CreateSender(url)
+
+	instanceID := "1"
+	instanceName := "super-cool-instance"
+
+	// Format your message according to the different channels to which it will be sent
+	text := fmt.Sprintf("Instance %s has exploded", instanceName)
+	html := fmt.Sprintf("Instance <b>%s</b> has exploded", instanceName)
+	markdown := fmt.Sprintf("Instance _%s_ has exploded", instanceName)
+
+	timestamp := time.Now()
+
+	message := notifications.Message{
+		Browser: notifications.BrowserMessage{
+			Type:      "critical",
+			Text:      text,
+			Timestamp: timestamp,
+		},
+		Email: notifications.EmailMessage{
+			Subject: "Explosion!",
+			Body:    html,
+		},
+		Slack: notifications.SlackMessage{
+			Text: markdown,
+		},
+	}
+
+	if err := sender.SendEvent("critical", instanceID, timestamp, &message); err != nil {
+		fmt.Printf("Error: %v", err)
+	}
+}
+```
--- a/notifications/notifications.go
+++ b/notifications/notifications.go
@@ -16,43 +16,44 @@ type Sender struct {
 }
 
 // EmailMessage contains the required fields for formatting email messages
-type EmailMessage struct {
+type emailMessage struct {
 	Subject string `json:"subject"`
 	Body    string `json:"body"`
 }
 
-type Attachment struct {
-	Fallback string   `json:"fallback,omitempty"`
-	Text     string   `json:"text"`
-	Color    string   `json:"color,omitempty"`
-	Markdown []string `json:"mrkdwn_in,omitempty"`
+type attachment struct {
+	Fallback   string   `json:"fallback,omitempty"`
+	Text       string   `json:"text"`
+	Color      string   `json:"color,omitempty"`
+	MarkdownIn []string `json:"mrkdwn_in,omitempty"`
 }
 
 // SlackMessage contains the required fields for formatting slack messages
-type SlackMessage struct {
+type slackMessage struct {
 	Text        string       `json:"text"`
-	Attachments []Attachment `json:"attachments"`
+	Attachments []attachment `json:"attachments"`
 }
 
 // BrowserMessage contains the required fields for formatting browser notifications
-type BrowserMessage struct {
+type browserMessage struct {
 	Type      string    `json:"type"`
 	Text      string    `json:"text"`
 	Timestamp time.Time `json:"timestamp"`
 }
 
 // Message contains the mappings for formatting notification messages
-type Message struct {
-	Browser BrowserMessage `json:"browser"`
-	Slack   SlackMessage   `json:"slack"`
-	Email   EmailMessage   `json:"email"`
+type message struct {
+	Browser  browserMessage `json:"browser"`
+	Slack    slackMessage   `json:"slack"`
+	Email    emailMessage   `json:"email"`
+	Fallback string         `json:"fallback"`
 }
 
 type event struct {
 	Type       string    `json:"type"`
 	InstanceID string    `json:"instance_id"`
 	Timestamp  time.Time `json:"timestamp"`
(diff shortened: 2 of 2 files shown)


notifications/notifications.go
```diff
@@ -0,0 +1,93 @@
...
+	e := event{
+		Type:       et,
+		InstanceID: instance,
+		Timestamp:  t,
+		Messages:   msg,
+	}
+
+	eventBytes, err := json.Marshal(e)
+	if err != nil {
+		return errors.Wrapf(err, "Cannot marshal event to []byte %v", msg)
```
lelenanam: here better to use `%s`. and print `e` not `msg`?

notifications/notifications.go
```diff
@@ -0,0 +1,93 @@
...
+
+// SlackMessage contains the required fields for formatting slack messages
+type SlackMessage struct {
+	Text string `json:"text"`
+}
+
+// BrowserMessage contains the required fields for formatting browser notifications
+type BrowserMessage struct {
+	Type      string    `json:"type"`
+	Text      string    `json:"text"`
```
lelenanam: Why both: text and message, for plain text and HTML?

notifications/notifications.go
```diff
@@ -0,0 +1,93 @@
...
+
+// EmailMessage contains the required fields for formatting email messages
+type EmailMessage struct {
+	Subject string `json:"subject"`
+	Body    string `json:"body"`
+}
+
+// SlackMessage contains the required fields for formatting slack messages
+type SlackMessage struct {
+	Text string `json:"text"`
```
lelenanam: should we add `"username": "Weave Cloud"`?

notifications/README.md
```diff
@@ -0,0 +1,49 @@
...
+			Text:      text,
+			Timestamp: timestamp,
+		},
+		Email: notifications.EmailMessage{
+			Subject: "Explosion!",
+			Body:    html,
+		},
+		Slack: notifications.SlackMessage{
+			Text: markdown,
+		},
```
lelenanam: What if a client wants to see `timestamp` and `type` in `slack` and `email` as well?

notifications/notifications.go
```diff
@@ -0,0 +1,93 @@
...
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return errors.Wrap(err, "Failed to POST event")
+	}
+
+	defer resp.Body.Close()
+
+	if err != nil {
```
lelenanam: should it be `err` from `close`? because you already checked `err` in line 82.

notifications/notifications.go
jpellizzari: I think the notification service itself should decide on what the username should be. This will be appearing in customer slack messages/emails, so it should be something "on-brand".

notifications/notifications.go
jpellizzari: Confused. The BrowserMessage does not have a `message` field

notifications/notifications.go
```diff
@@ -0,0 +1,93 @@
...
+
+// Message contains the mappings for formatting notification messages
+type Message struct {
+	Browser BrowserMessage `json:"browser"`
+	Slack   SlackMessage   `json:"slack"`
+	Email   EmailMessage   `json:"email"`
+}
+
+type event struct {
+	Type       string    `json:"type"`
```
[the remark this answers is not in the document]
jpellizzari: The client will not know the ID. Eventmanager will assign an ID on creation.

notifications/README.md
jpellizzari: They have to embed that in the message body. The browser has special logic for handling different event types and timestamps. 

The equivalent would be to specify HTML with the correct inline styling or class names in the message in order for the notification to be styled correctly. That puts too much strain/work on the user of this API.

notifications/notifications.go
lelenanam: https://github.com/weaveworks/notification/issues/43

notifications/notifications.go
lelenanam: I mean `Message` from line 33 here

notifications/notifications.go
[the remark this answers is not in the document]
lelenanam: ok https://github.com/weaveworks/notification/issues/44

notifications/notifications.go
jpellizzari: Oh whoops! Nice catch

notifications/notifications.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "POST request error to URL %s", s.URL)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return errors.Wrap(err, "Failed to POST event")
+	}
```
jml: Ideally we'd have client-side Prometheus instrumentation here, so we can see requests, response codes, and latency in our metrics. Having this early on can actually make development a lot easier too.

I had a quick poke around for examples but couldn't find any. The `users` client in service is close, but it's gRPC (we went through a phase of trying to use gRPC for internal APIs).

jpellizzari: @jml I'm not seeing any prom instrumentation on the users client in weaveworks/service. It seems odd to instrument both the client and the server.

Can you link me to the example you mentioned?

jml: My mistake. The billing version of the users client has such instrumentation: https://github.com/weaveworks/billing/blob/master/users/client.go

Complicated by 
a) billing shouldn't have a copy, it should use the users client
b) that instrumentation should, IMO, be in the users client itself

Likewise, the cortex configs service client (which I wrote!) gets this wrong. Users of the client do their own instrumentation (e.g. https://github.com/weaveworks/cortex/blob/master/pkg/ruler/scheduler.go#L36 – follow the `configsRequestDuration` variable), rather than the client itself providing it.

notifications/notifications.go
```diff
@@ -0,0 +1,88 @@
...
+
+// CreateSender creates a Sender
+func CreateSender(url string) Sender {
+	return Sender{
+		URL: url,
+	}
+}
+
+// SendEvent sends an event to the notification service.
+func (s *Sender) SendEvent(et string, instance string, t time.Time, msg *Message) error {
```
ekimekim: This being a public api, it's not clear what `et` is. It should be spelt out in the comment above, or at the very least spelt out in full (`eventType` instead of `et`).

notifications/notifications.go
```diff
@@ -0,0 +1,88 @@
...
+	Browser BrowserMessage `json:"browser"`
+	Slack   SlackMessage   `json:"slack"`
+	Email   EmailMessage   `json:"email"`
+}
+
+type event struct {
+	Type       string    `json:"type"`
+	InstanceID string    `json:"instance_id"`
+	Timestamp  time.Time `json:"timestamp"`
+	Messages   *Message  `json:"messages"`
```
bboreham: Why is this a pointer?  What does it mean for it to be nil?

notifications/notifications.go
```diff
@@ -0,0 +1,127 @@
...
+	e := event{
+		Type:       eventType,
+		InstanceID: instance,
+		Timestamp:  t,
+		Messages:   createMessage(eventType, t, msg, attachments),
+	}
+
+	eventBytes, err := json.Marshal(e)
+	if err != nil {
+		return errors.Wrapf(err, "Cannot marshal event to []byte %s", err)
```
rndstr: nit: since this is a library I think the Go way would be to have these lowercase and prefix with package name, such as `notifications: cannot marshal …`

PR #63, opened 2017-09-22
If orgID already present, rename instead of reject

leth:
As discussed in a related private change, I am trying a certain request from an internal service which results in this header already being present.

If we simply rename the offending header, rather than reject the request, then things should work.

### DIFF
--- a/user/grpc.go
+++ b/user/grpc.go
@@ -33,19 +33,14 @@ func InjectIntoGRPCRequest(ctx context.Context) (context.Context, error) {
 		md = metadata.New(map[string]string{})
 	}
 	newCtx := ctx
+	md = md.Copy()
 	if orgIDs, ok := md[lowerOrgIDHeaderName]; ok {
-		if len(orgIDs) == 1 {
-			if orgIDs[0] != orgID {
-				return ctx, ErrDifferentOrgIDPresent
-			}
-		} else {
-			return ctx, ErrTooManyOrgIDs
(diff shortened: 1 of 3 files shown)


rade:
Why rename it? What is looking at the new "orig" headers introduced in this PR?

leth:
True, I don't plan to look at it. I was mostly thinking it might be useful to have it around for debugging weird issues. I also considered logging a warning or debug message.

Happy to change it, what do people prefer?

user/grpc.go
```diff
@@ -33,19 +33,14 @@ func InjectIntoGRPCRequest(ctx context.Context) (context.Context, error) {
...
+		if len(orgIDs) != 1 || orgIDs[0] != orgID {
+			md[lowerOrigOrgIDHeaderName] = orgIDs
 		}
-	} else {
-		md = md.Copy()
-		md[lowerOrgIDHeaderName] = []string{orgID}
-		newCtx = metadata.NewOutgoingContext(ctx, md)
 	}
+	md[lowerOrgIDHeaderName] = []string{orgID}
+	newCtx = metadata.NewOutgoingContext(ctx, md)
```
rndstr: you could directly return here and also get rid of the `newCtx` var altogether

leth:
Going to decline this, in favor of adding a middleware to authfe which drops any offending incoming headers.
Changing it here has complicated implications on other services that use this, and the checks may have some safety/sanity benefit to them.

PR #64, opened 2017-10-02
Header names are public members

leth:
To avoid copying the names into other modules.


leth:
Argh gocycle has become more picky.

leth:
Nope turns out the linter didn't like my comments.

PR #65, opened 2017-10-03
Add Prometheus metrics for number of log messages

marccarre:
[Promrus](https://github.com/marccarre/promrus/) is a Logrus hook to expose the number of log messages as [Prometheus](https://prometheus.io/) metrics:
```
log_messages{level="debug"}
log_messages{level="info"}
log_messages{level="warning"}
log_messages{level="error"}
```

Added dependency using:
```
$ dep ensure github.com/weaveworks/promrus@v1.0.0 && dep ensure && dep prune
```

Fixed casing in `logrus` imports, causing the following issue:
```
middleware/grpc_logging.go:10:2: case-insensitive import collision:
"github.com/weaveworks/common/vendor/github.com/Sirupsen/logrus" and
"github.com/weaveworks/common/vendor/github.com/sirupsen/logrus"
```
by:
- changing `import` statements from `Sirupsen/logrus` to `sirupsen/logrus`, as per [the official documentation](https://github.com/sirupsen/logrus), and
- running `dep ensure && dep prune`.

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -10,9 +10,9 @@ import (
 
 	"golang.org/x/net/context"
 
-	"github.com/marccarre/promrus"
 	log "github.com/sirupsen/logrus"
 	"github.com/weaveworks/common/user"
+	"github.com/weaveworks/promrus"
 )
 
 const (
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -10,8 +10,8 @@ import (
 
 	"golang.org/x/net/context"
 
-	log "github.com/Sirupsen/logrus"
 	"github.com/marccarre/promrus"
+	log "github.com/sirupsen/logrus"
 	"github.com/weaveworks/common/user"
 )
 
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -11,6 +11,7 @@ import (
 	"golang.org/x/net/context"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/marccarre/promrus"
 	"github.com/weaveworks/common/user"
 )
 
@@ -27,6 +28,11 @@ func Setup(logLevel string) error {
 	}
 	log.SetLevel(level)
 	log.SetFormatter(&textFormatter{})
+	hook, err := promrus.NewPrometheusHook() // Expose number of log messages as Prometheus metrics.
+	if err != nil {
+		return err
+	}
+	log.AddHook(hook)
 	return nil
 }
 
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -116,12 +116,6 @@
   revision = "b84e30acd515aadc4b783ad4ff83aff3299bdfe0"
 
 [[projects]]
-  name = "github.com/marccarre/promrus"
-  packages = ["."]
-  revision = "6f9713cf3e9bc69424a25d8d0f696f7ce7b4ff2b"
-  version = "v1.0.0"
-
-[[projects]]
   name = "github.com/mattn/go-colorable"
   packages = ["."]
   revision = "d228849504861217f796da67fae4f6e347643f15"
@@ -242,6 +236,12 @@
   revision = "17ff1516db3dffe25452b4cbf6e6412f303cddea"
 
 [[projects]]
+  name = "github.com/weaveworks/promrus"
+  packages = ["."]
+  revision = "6f9713cf3e9bc69424a25d8d0f696f7ce7b4ff2b"
+  version = "v1.0.0"
+
+[[projects]]
   branch = "master"
   name = "golang.org/x/crypto"
   packages = ["ssh/terminal"]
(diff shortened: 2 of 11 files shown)


marccarre:
What do you mean by not sure? 
Do you think it could cause issues to change it? 
Is there some sort of a convention I missed regarding this?

I initially did not realise this would lead to such changes when I added `sirupsen/logrus` to my dependencies in `promrus` by simply following their documentation. I then realised there was this casing difference here, but also just saw, for example, that Cortex's dependencies are also using `Sirupsen/logrus`, so if I keep `promrus` as is, I believe it might lead to more build issues, case changes (I don't mind making these changes if they are painless), or even preventing metrics from being exposed since Go might treat them as different packages altogether (?).
Alternatively, I could just change `promrus` to use `Sirupsen/logrus` instead of `sirupsen/logrus`.

marccarre:
In any case, I'll try it against Cortex and will report back.

jml:
Not sure as in "I did something like this a few months ago and regretted it but can't recall the details". It might well break things.

marccarre:
FYI, looks like we aren't the only ones facing this issue, see [here](https://github.com/sirupsen/logrus/issues/543#issuecomment-306585115) and [here](https://github.com/sirupsen/logrus/issues/570#issuecomment-313933276), and that the way forward is to use lowercase. I'll bite the bullet and make the necessary changes where we decide to use `promrus`.

marccarre:
I haven't tried on Cortex yet, but trying on a smaller project mimicking its dependencies shows that:
1. having mixed casing in `vendor` definitely breaks the build,
2. even if you only have `sirupsen/logrus` in `vendor` (e.g.: `dep ensure` and then `rm -fr vendor/github.com/Sirupsen`), instrumenting on `sirupsen/logrus` does not instrument `Sirupsen/logrus`:

`foo.go` in project `foo`:
```
package foo

import (
	log "github.com/Sirupsen/logrus"
)

// Sirupsen sirupsens.
func Sirupsen() {
	log.Infof("Sirupsen/logrus")
}
```
`main.go` in a different project:
```
package main

import (
	"net/http"
	"time"

	"github.com/marccarre/promrus"
	"github.com/marccarre/foo"
	"github.com/prometheus/client_golang/prometheus/promhttp"
	log "github.com/sirupsen/logrus"
)

func main() {
	hook, err := promrus.NewPrometheusHook()
	if err != nil {
		return
	}
	log.AddHook(hook)
	go http.ListenAndServe(":8080", promhttp.Handler())
	for {
		foo.Sirupsen()
		time.Sleep(10 * time.Second)
	}
}
```
yields metrics which do not increase over time:
```
$ curl -fsS localhost:8080 | grep log_messages
# HELP log_messages Total number of log messages.
# TYPE log_messages counter
log_messages{level="debug"} 0
log_messages{level="error"} 0
log_messages{level="info"} 0
log_messages{level="warning"} 0
```

bboreham:
Why do we even want this?

marccarre:
What do you mean by "this"?
- Adding metrics was suggested on this issue: https://github.com/weaveworks/service-conf/issues/977#issuecomment-333528506 
- Lower-casing was suggested on this issue: https://github.com/sirupsen/logrus/issues/570#issuecomment-313933276

Feel free to revert if causing issues though.

logging/logging.go
```diff
@@ -27,6 +28,11 @@ func Setup(logLevel string) error {
 	}
 	log.SetLevel(level)
 	log.SetFormatter(&textFormatter{})
+	hook, err := promrus.NewPrometheusHook() // Expose number of log messages as Prometheus metrics.
```
bboreham: Please add to the documentation on this function that it cannot be called more than once, or bracket this section with a test to avoid causing an `error duplicate metrics collector registration attempted` (or adding multiple instances of the same hook below)

marccarre: Another solution could be to unregister the counter before registering it, in which case initialising multiple times will not cause any issue:
```
func NewPrometheusHook() (*PrometheusHook, error) {
[...]
	// Try to unregister the counter vector, in case already registered for some reason,
	// e.g. double initialisation/configuration done by mistake by the end-user.
	prometheus.Unregister(counterVec)
	// Try to register the counter vector:
	err := prometheus.Register(counterVec)
[...]
}
```
Test:
```
func TestRegisteringHookMultipleTimesShouldBeSafe(t *testing.T) {
	log.AddHook(promrus.MustNewPrometheusHook())
	log.AddHook(promrus.MustNewPrometheusHook()) // Initialising twice in a row should NOT raise any error.

	httpServePrometheusMetrics(t)

	lines := httpGetMetrics(t)
	assert.Equal(t, 0, countFor(t, log.InfoLevel, lines))

	log.Info("this is at info level!")

	lines = httpGetMetrics(t)
	assert.Equal(t, 1, countFor(t, log.InfoLevel, lines))

	log.AddHook(promrus.MustNewPrometheusHook()) // Initialising again here should NOT raise any error, but will reset counters.

	lines = httpGetMetrics(t)
	assert.Equal(t, 0, countFor(t, log.InfoLevel, lines))
}
```
@bboreham, do you think this could be a better solution?

marccarre: Since we should prob. not discuss this here, issue opened weaveworks/promrus#5 and PR ready for review weaveworks/promrus#6.

PR #70, opened 2017-10-10
Upgrade promrus from v1.0.0 to v1.2.0.

marccarre:


marccarre:
Merged directly since nothing really worth reviewing, and upgrading shouldn't break anything downstream.

PR #72, opened 2017-10-16
Allow nil histograms in instrumentation

juliusv:


juliusv:
@jml @rndstr Cortex needs to upgrade to the latest `weaveworks/common`, but the current one doesn't allow instrumentation where the histogram is `nil` (to only do tracing). This adds back that capability, so we can continue using it from Cortex.

PR #73, opened 2017-10-17
Add timed client request taking a collector

rndstr:
New method is named `client.TimeRequest()`. This avoids
recreation of the collector on every request.

Also adds `instrument.HistogramCollectorFromOpts()` and
makes the interface `instrument.Collector` public.

### DIFF
--- a/http/client/client.go
+++ b/http/client/client.go
@@ -16,8 +16,8 @@ type Requester interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
-// TimeRequestHistogram performs an HTTP client request and records the duration in a histogram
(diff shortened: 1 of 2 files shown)


http/client/client.go
```diff
@@ -30,8 +30,14 @@ func TimeRequestHistogram(ctx context.Context, operation string, metric *prometh
...
 		return "error"
 	}
-	coll := instrument.NewHistogramCollector(metric)
 	err := instrument.CollectedRequest(ctx, fmt.Sprintf("%s %s", request.Method, operation),
 		coll, toStatusCode, doRequest)
 	return response, err
 }
+
+// TimeRequestHistogram performs an HTTP client request and records the duration in a histogram.
+// Try to use TimeRequest() to avoid creation of a collector on every request.
```
leth: Should we use the deprecation convention here? `// Deprecated: reason`

PR #74, opened 2017-10-18
Add `client.TimedClient` implementing `Requester`

rndstr:
This helps in chaining clients doing requests. Such as

    client.NewTimedClient(http.DefaultClient, someCollector}

### DIFF
--- a/http/client/client.go
+++ b/http/client/client.go
@@ -16,6 +16,33 @@ type Requester interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
+// TimedClient instruments a request. It implements Requester.
+type TimedClient struct {
+	client    Requester
+	collector instrument.Collector
+}
+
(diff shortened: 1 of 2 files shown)


http/client/client.go
```diff
@@ -16,6 +16,33 @@ type Requester interface {
...
 
+// TimedClient instruments a request. It implements Requester.
+type TimedClient struct {
+	client    Requester
+	collector instrument.Collector
+}
+
+// CtxTimedOperationNameKey specifies the operation name location within the context
+// for instrumentation.
+const CtxTimedOperationNameKey = "op"
```
leth: Will this cause problems if two timing contexts are nested?

rndstr: Shouldn't, because the context is tied to the request. (If you nest `TimedClient`s it will just measure it twice for the same op)

PR #77, opened 2017-11-20
Add jaeger initialization

rbruggem:


tracing/tracing.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+
+	jaeger "github.com/uber/jaeger-client-go"
+	jaegercfg "github.com/uber/jaeger-client-go/config"
+)
+
+type nopCloser struct {
```
rndstr: FWIW there is `ioutil.NopCloser`

tracing/tracing.go
```diff
@@ -0,0 +1,40 @@
...
+				LocalAgentHostPort: fmt.Sprintf("%s:6831", jaegerAgentHost),
+			},
+		}
+
+		closer, err := cfg.InitGlobalTracer(serviceName)
+		if err != nil {
+			fmt.Printf("Could not initialize jaeger tracer: %s\n", err.Error())
+			os.Exit(1)
+		}
+		return closer
```
jml: Do you think there's any value in extracting the bit inside this condition such that there's a function that takes a `jaegerAgentHost` and a `serviceName` and returns a tracer?

tracing/tracing.go
rbruggem: thanks!

tomwilkie:
Do you think its worth removing the dependancies on Loki from server.go?  https://github.com/weaveworks/common/blob/79ec4e6e7a299b05f8e6a78912251cfc0432602c/server/server.go#L29

That will make it easier for different projects to use these libraries and not pull in the massive rats nest of dependancies the Loki entails.  They can always pull in Loki themselves if needs be.

rbruggem:
@tomwilkie yes, but I believe it should be in a different PR.

PR #78, opened 2017-11-21
Add error message to OpenTracing span

bboreham:
This is handy when debugging.  ~I limit the length arbitrarily just in case someone has dumped their life-story into the error string.~

### DIFF
--- a/instrument/instrument.go
+++ b/instrument/instrument.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
+	otlog "github.com/opentracing/opentracing-go/log"
 	"github.com/prometheus/client_golang/prometheus"
 	oldcontext "golang.org/x/net/context"
 )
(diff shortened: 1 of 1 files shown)


instrument/instrument.go
```diff
@@ -140,6 +140,12 @@ func CollectedRequest(ctx context.Context, method string, col Collector, toStatu
 
 	if err != nil {
 		ext.Error.Set(sp, true)
+		errStr := err.Error()
+		const maxTagErrorLen = 128
```
jml: I'd paint it 1024. I figure we should allow tweet-length messages at a minimum.

bboreham: 140? 280?

jml: Whichever is more presidential.

tomwilkie:
I think the OpenTracing "way" is to do `span.LogFields(log.Error(err))`: https://github.com/opentracing-contrib/go-stdlib/pull/14#discussion_r147025367

bboreham:
Changed as suggested - thanks @tomwilkie !
(Not limiting the length now on the expectation that OpenTracing logs are better able to cope with long strings than tags)

PR #86, opened 2018-02-07
Update prom client go

dlespiau:
By default, go dep will take the latest stable version released. For client_golang this means it will lock the 0.8.0 version from 2016-08-17. It seems a sub-optimal default and was causing our golang services to be missing interesting metrics and bug fixes.
    
Also filed: https://github.com/prometheus/client_golang/issues/375

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -21,7 +21,13 @@
 
 [[projects]]
   name = "github.com/aws/aws-sdk-go"
-  packages = ["aws","aws/awserr","aws/credentials","aws/endpoints","internal/shareddefaults"]
+  packages = [
+    "aws",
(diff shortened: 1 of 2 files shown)


aaron7:
Should we pin to an exact version instead of latest master to get more reproducable builds?

dlespiau:
The pinning is done by the lock file: you'll always use the hash from the lock file (and so do reproducible builds) until the next `go dep ensure -update $package` that will pick up the current HEAD on master.

aaron7:
> The pinning is done by the lock file: you'll always use the hash from the lock file (and so do reproducible builds) until the next go dep ensure -update $package that will pick up the current HEAD on master.

Ah, of course. Thanks

PR #87, opened 2018-02-13
Add args to tracing.New to allow configurable trace sampling

cboggs:
This commit allows Cortex systems to specify, per-service, how trace sampling should be handled in the Jaeger client. Given that `jaeger.SamplerTypeConst` yields an all-or-nothing approach to sampling, Production deployments of Cortex are effectively blocked from enabling Jaeger traces. The load generated on Jaeger Collector (and its data store) is immense with a sampling probability of 1, even for moderate-sized deployments.

There's a partner PR (https://github.com/weaveworks/cortex/pull/703) to enable Cortex components to take advantage of this, defaulting to `jaeger.SamplerTypeRateLimiting` with a param of `10.0` to yield ~10 traces per second being shipped when `JAEGER_AGENT_HOST` is set at runtime. This is easily swapped for an alternate sampling strategy, however.

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -5,23 +5,30 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"strconv"
 
 	jaegercfg "github.com/uber/jaeger-client-go/config"
 )
 
 // New registers Jaeger as the OpenTracing implementation.
 // If jaegerAgentHost is an empty string, tracing is disabled.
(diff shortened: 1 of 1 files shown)


jml:
Thank you!

cboggs:
Sure thing! Thanks for the feedback. I went back and forth a bit in my head about that very approach, so the confirmation was great. :-)

PR #89, opened 2018-03-06
Don't log full request body on failed /cortex.Ingester/Push calls

cboggs:
#85 added across-the-board logging of gRPC body contents on failed requests, which blows up the Ingester logs on a failed push.

In my perusal so far, ingester push failures currently log either:
* Nothing at all about the relevant series
* All the pertinent details such as metric name, label values, latest value, incoming value, etc.

This change should simply retain the old logging behavior for ingester push failures, while still enhancing logging of other RPC failures across the system.

@bboreham, I think this may unblock https://github.com/weaveworks/cortex/pull/709 (once vendored in of course). What are your thoughts?

### DIFF
--- a/middleware/grpc_logging.go
+++ b/middleware/grpc_logging.go
@@ -18,7 +18,11 @@ var ServerLoggingInterceptor = func(ctx context.Context, req interface{}, info *
 	resp, err := handler(ctx, req)
 	entry := logging.With(ctx).WithFields(log.Fields{"method": info.FullMethod, "duration": time.Since(begin)})
 	if err != nil {
-		entry.WithField("request", req).WithError(err).Warn(gRPC)
+		if info.FullMethod == "/cortex.Ingester/Push" {
+			entry.WithError(err).Warn(gRPC)
+		} else {
+			entry.WithField("request", req).WithError(err).Warn(gRPC)
+		}
 	} else {
 		entry.Debugf("%s (success)", gRPC)
 	}
--- a/middleware/grpc_logging.go
+++ b/middleware/grpc_logging.go
@@ -33,6 +33,3 @@ func (s GRPCServerLog) UnaryServerInterceptor(ctx context.Context, req interface
 	}
 	return resp, err
 }
-
-// ServerLoggingInterceptor logs gRPC requests, errors and latency. Left here for backwards compatibility
-var ServerLoggingInterceptor = GRPCServerLog{WithRequest: true}.UnaryServerInterceptor
--- a/middleware/grpc_logging.go
+++ b/middleware/grpc_logging.go
@@ -12,29 +12,27 @@ import (
 
 const gRPC = "gRPC"
 
-// ServerLoggingInterceptor logs gRPC requests, errors and latency.
-var ServerLoggingInterceptor = func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-	begin := time.Now()
-	resp, err := handler(ctx, req)
-	entry := logging.With(ctx).WithFields(log.Fields{"method": info.FullMethod, "duration": time.Since(begin)})
-	if err != nil {
-		entry.WithField("request", req).WithError(err).Warn(gRPC)
-	} else {
-		entry.Debugf("%s (success)", gRPC)
-	}
-	return resp, err
+// GRPCServerLog logs grpc requests, errors, and latency.
+type GRPCServerLog struct {
+	// WithRequest will log the entire request rather than just the error
+	WithRequest bool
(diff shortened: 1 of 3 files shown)


cboggs:
@marchbanks has pointed out a remarkably cleaner approach to this. Will fix and push shortly. :-)

cboggs:
Much better, I think. :-)
I've tested this in our staging cluster, adding the requisite boolean to the Ingester's `server.New` call to make it use the new truncated log. I forced many many out-of-order timestamps, and this is the output:

```
ingester-58b6b74757-cbfd4 ingester time="2018-03-06T17:18:36Z" level=warning msg=gRPC duration=36.668857ms error="rpc error: code = Code(400) desc = sample timestamp out of order for series kube_replicaset_status_replicas{<snipped for brevity>}; last timestamp: 1520356706.133, incoming timestamp: 1520356703.917" method=/cortex.Ingester/Push
```

server/server.go
```diff
@@ -92,6 +93,8 @@ func New(cfg Config) (*Server, error) {
 		return nil, err
 	}
 
+	var loggingInterceptor func(context.Context, interface{}, *grpc.UnaryServerInfo, grpc.UnaryHandler) (interface{}, error)
```
bboreham: This would be simpler as:
```
loggingInterceptor := middleware.ServerLoggingInterceptor
if cfg.TruncatedServerLogging {
	loggingInterceptor = middleware.TruncatedServerLoggingInterceptor
}
```

middleware/grpc_logging.go
```diff
@@ -24,3 +24,17 @@ var ServerLoggingInterceptor = func(ctx context.Context, req interface{}, info *
 	}
 	return resp, err
 }
+
+// TruncatedServerLoggingInterceptor logs gRPC requests, errors and latency,
+// but omits request bodies.
+var TruncatedServerLoggingInterceptor = func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
```
bboreham: "Truncated" suggests to me that it will keep the first N bytes and omit the rest.  

How about making this a member function on an object with `withRequest` bool field?

cboggs: Sorry to be slow, but I'm not sure where best to put said member function?

bboreham: In this file.  So what are currently `ServerLoggingInterceptor` and `TruncatedServerLoggingInterceptor` would be just one member on an object with the config.

Something like `Log` at https://github.com/weaveworks/common/blob/master/middleware/logging.go#L15-L17

csmarchbanks:
@bboreham I updated this commit based on your suggestions, definitely like it more. Do you know if middleware.ServerLoggingInterceptor is used anywhere outside of weaveworks/common? I.e. is it safe to delete?

bboreham:
`ServerLoggingInterceptor` only shows up in `common` when I do a search across all Weaveworks repos.
(and I believe I have access to all of them)

middleware/grpc_logging.go
```diff
@@ -12,15 +12,26 @@ import (
...
-		entry.Debugf("%s (success)", gRPC)
+// GRPCServerLog logs grpc requests, errors, and latency.
+type GRPCServerLog struct {
+	// WithRequest will log the entire request rather than just the error
+	WithRequest bool
+}
+
+// UnaryServerInterceptor returns an interceptor that logs gRPC requests
+func (s GRPCServerLog) UnaryServerInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
```
bboreham: You don't have to do this as `return func` - just declare a method with the right signature and Go will create a closure for it if you refer to it as `serverLog.UnaryServerInterceptor`
https://golang.org/ref/spec#Method_values

csmarchbanks:
Ok, I removed `ServerLoggingInterceptor` if someone runs into issues it will be easy to add it back in.

server/server.go
```diff
@@ -42,6 +42,7 @@ type Config struct {
 	GRPCListenPort   int
 
 	RegisterInstrumentation bool
+	TruncatedServerLogging  bool
```
bboreham: could name this after what it does, e.g. `LogRequest` or `IncludeRequestInLog` or something like that

middleware/grpc_logging.go
csmarchbanks: That's neat!

csmarchbanks:
I renamed the config variable to "ExcludeRequestInLog" for backwards compatibility.

PR #92, opened 2018-04-02
Return errors when http/grpc servers exit.

gouthamve:
There was a case where the HTTP server died but the process exited. We
need to exit when the servers exit.

Quirky that there are unused imports before, what am I missing?

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/common/92)


gouthamve:
Didn't run `dep ensure` and go-imports automatically removed the imports. Explicitly added the alias so that that doesn't happen again.

server/server.go
```diff
@@ -6,19 +6,20 @@ import (
 	"net"
 	"net/http"
 	_ "net/http/pprof" // anonymous import to get the pprof handler registered
+	"os"
 	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
-	"github.com/mwitkow/go-grpc-middleware"
+	grpc_middleware "github.com/mwitkow/go-grpc-middleware"
```
bboreham: Please make a separate PR to make unrelated changes.

server/server.go
```diff
@@ -162,12 +163,22 @@ func RegisterInstrumentation(router *mux.Router) {
 
 // Run the server; blocks until SIGTERM is received.
 func (s *Server) Run() {
-	go s.httpServer.Serve(s.httpListener)
+	go func() {
+		if err := s.httpServer.Serve(s.httpListener); err != http.ErrServerClosed {
+			log.Errorf("error serving http: %v", err)
+			os.Exit(1)
```
bboreham: I don't think we can call `os.Exit()` unconditionally, from a library.

bboreham:
Can you provide a bit more context on what went wrong?

bboreham:
Maybe change it so that `Run()` returns with an error if one of the serving loops does?

bboreham:
@gouthamve are you likely to come back to this?

gouthamve:
Oh, I will sometime this week. Just added it back to my list. Sorry.

gouthamve:
Extremely sorry for the delay. So what I observed was that the http server shutdown but the service was still running. After exec'ing in, I saw that the port was not bound essentially saying the http server exited.

I think we should catch those errors and not just ignore them.

bboreham:
Looking much better.  Any chance of a test?

gouthamve:
@bboreham Done. Please take a look.

server/server.go
```diff
@@ -161,17 +161,39 @@ func RegisterInstrumentation(router *mux.Router) {
...
+
+	go func() {
+		if err := s.httpServer.Serve(s.httpListener); err != http.ErrServerClosed {
+			select {
+			case errChan <- err:
+			default:
+			}
+		}
+	}()
+	defer s.httpServer.Close()
```
tomwilkie: As per #99, we shouldn't shutdown the httpServer here.

tomwilkie:
Just seen this again (HTTP server randomly stopping serving traffic).

server/server.go
```diff
@@ -161,17 +161,38 @@ func RegisterInstrumentation(router *mux.Router) {
 }
 
 // Run the server; blocks until SIGTERM is received.
-func (s *Server) Run() {
-	go s.httpServer.Serve(s.httpListener)
+func (s *Server) Run() error {
```
bboreham: I think the comment should say "... or an error is returned".

gouthamve:
You're right. Nice, catch. I've made sure the function now returns. 

I noticed a race-condition now.... So if we send a SIGTERM, or an error is returned, right after calling `Run()`, the function never exits. This is the reason there is a `Sleep` in the tests, I now see that it can cause some problems.

For example, if the HTTP server has a closed connection, it'll immediately exit, but by that time, we might not have evaluated `return <-errChan`, hence the error will fall into the default section. And `Run()` will never exit.

Seems unpractically rare (tests failing w/o sleep point otherwise), but curious about your thoughts.

bboreham:
I can't see how an error return would not result in exiting the function.
SIGTERM could arrive before we started listening for it, and if we're pid 1 would do nothing.  So move the signal listen up to the top of the function to reduce the window.

bboreham:
I still have to look into your race condition, but there is a merge conflict you might like to fix in the meantime.

bboreham:
I can't run `TestRunReturnsError` in my standard config because both sub-tests run in parallel and they both use the same ports.

bboreham:
Running them one at a time, I can't get it to fail by taking the sleep out.

bboreham:
I cannot replicate the race condition you reported, but I do find that Go can run the two `TestRunReturnsError` in parallel and hence fail on port in use.

gouthamve:
@bboreham Done. And made sure the test doesn't fail when run.

PR #95, opened 2018-05-16
Add gRPC streaming middleware.

tomwilkie:
Adds a series of gRPC _stream_ middleware for logging, user header propagation and tracing.  This is laying groundwork for using gRPC streaming between ingesters and queriers in Cortex.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -6,3 +6,7 @@
 [[constraint]]
   name = "github.com/prometheus/client_golang"
   branch = "master"
+
+[[constraint]]
+  name = "github.com/sercand/kuberesolver"
+  revision = "aa801ca262949d887bbe0bae3f6f731ac82c26f6"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -7,6 +7,9 @@
   name = "github.com/prometheus/client_golang"
   branch = "master"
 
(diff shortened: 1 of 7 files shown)


Gopkg.toml
```diff
@@ -6,3 +6,7 @@
 [[constraint]]
   name = "github.com/prometheus/client_golang"
   branch = "master"
+
+[[constraint]]
```
bboreham: When adding a constraint or override, please include a comment giving the reason, and the conditions under which it can be removed.

server/server.go
```diff
@@ -57,7 +58,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 
 // Server wraps a HTTP and gRPC server, and some common initialization.
 //
-// Servers will be automatically instrumented for Prometheus metrics.
+// Servers will be automatically instrumented for Prometheus metrics
+// and Loki tracing.  HTTP over gRPC
```
bboreham: surely "whatever OpenTracing you have configured", not necessarily Loki.

tomwilkie: Sorry bad merge there I think; that shouldn't be there.

tomwilkie:
@bboreham thanks! PTAL

PR #96, opened 2018-06-01
Abstract logging for the middleware

tomwilkie:
Idea here is for an abstract Interface that 'unifies' gokit logging and logrus logging, such that
the middleware in this repo can be used in projects which use either loggers.

Once this exists, we can also dedupe some of the logging "middleware" we've got, for instance the ones that count log lines at various levels.  Thats still todo, in another PR.


logging/dedupe.go
```diff
@@ -9,6 +9,43 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const (
```
bboreham: seems this change should be part of the preceding commit.

tomwilkie: Fixed!  Lots of rebasing to try and make this all a little easier to review.

PR #99, opened 2018-07-13
Don't shutdown the gRPC server the minute a signal is received.

tomwilkie:
gRPC behaviour should be the same as HTTP: `Run()` exits when a signal is received, and `Stop()` stops the handling of requests.

See https://github.com/weaveworks/cortex/issues/882


tomwilkie:
Deployed to dev & tested, fixes the problem

bboreham:
Could you put a bit more explanation in the commit message please?  In particular the change to `Shutdown` doesn't seem to be covered at all, but what you put in the PR description also seems relevant.

PR #100, opened 2018-07-16
Add user and org ID to tracing span if known

bboreham:
This should make it easier to do things like filter on an instance.


bboreham:
This is somewhat duplicative of https://github.com/weaveworks/service/pull/1995, perhaps because I had forgotten about it.

bboreham:
Rebased and changed the tag names to match what I had used elsewhere.

It is still valid to set them here, as the other place is only for requests that pass through front-end authorization.

PR #102, opened 2018-07-16
When trace reporting is enabled, don't enable trace sampling unless asked

### LINKED ISSUE #101: Trace sampling should not be on by default when trace reporting is configured
Trace sampling should not be enabled by default if trace reporting is configured. It should be enabled separately

The decision about whether to capture a trace should be made by one service, as close to the request source as possible.
Then that decision should be passed along with the request and honored by each service involved in the request processing.

For example in cortex, if a service is configured to send samples to jaeger, it also defaults to a sampling rate of 10 per second.
If both the auth layer, distributor and ingester were configured thusly, you would see 10-20 traces per second from the distributor, and 20-30 traces per second from the ingester. Additionally you'd see some traces which start at the distributor or the ingester.

Cortex issue: https://github.com/weaveworks/cortex/issues/885

leth:
Reporting trace spans for a request is a separate task from choosing which requests should be traced.

Fixes #101

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -9,8 +9,8 @@ import (
 	jaegercfg "github.com/uber/jaeger-client-go/config"
 )
 
-// InstallJaeger registers Jaeger as the OpenTracing implementation.
-func InstallJaeger(serviceName string, cfg *jaegercfg.Configuration) io.Closer {
+// installJaeger registers Jaeger as the OpenTracing implementation.
+func installJaeger(serviceName string, cfg *jaegercfg.Configuration) io.Closer {
 	closer, err := cfg.InitGlobalTracer(serviceName)
 	if err != nil {
 		fmt.Printf("Could not initialize jaeger tracer: %s\n", err.Error())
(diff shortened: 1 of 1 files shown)


tracing/tracing.go
```diff
@@ -5,46 +5,37 @@ import (
...
+		fmt.Printf("Could not initialize jaeger tracer: %s\n", err.Error())
+		os.Exit(1)
 	}
-	return ioutil.NopCloser(nil)
+	return closer
 }
 
 // NewFromEnv is a convenience function to allow tracing configuration
 // via environment variables
+// Tracing is disabled unless one of the following environment variables is used to configure jaeger:
```
LiliC: So here we say unless one of the following... but below on line 35 we check that all 3 must not be empty. I think this needs rephrasing to match the code.

tracing/tracing.go
```diff
@@ -5,46 +5,37 @@ import (
...
-		}
-
-		closer, err := cfg.InitGlobalTracer(serviceName)
-		if err != nil {
-			fmt.Printf("Could not initialize jaeger tracer: %s\n", err.Error())
-			os.Exit(1)
-		}
-		return closer
+// InstallJaeger registers Jaeger as the OpenTracing implementation.
+func InstallJaeger(serviceName string, cfg *jaegercfg.Configuration) io.Closer {
```
LiliC: As this is an exported function, I am wondering are we sure we do not use this, and if its only used here can we just not export it?

leth: Yeah YAGNI I guess, we don't use it, I was mostly trying to make up for removing the `New` method.

tracing/tracing.go
leth: 🤦‍♂️ good catch

leth: Actually I think they do match up. I'll reword the comment to cancel out some negatives.

PR #109, opened 2018-07-25
Raise default server timeouts from 5s to 30s

bboreham:
Primary motivation is to allow a standard Go profiling request to complete, for which I only need the write timeout, but I raised the other 5-second timeouts too since it isn't that unusual to have operations or data uploads take longer.


tomwilkie:
LGTM, we have them raised on a lot of our jobs via flags.

PR #111, opened 2018-07-25
Allow the AWS backend to get credentials from its environment.

tomwilkie:
See https://github.com/weaveworks/cortex/pull/898

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -18,14 +18,7 @@ import (
 // endpoint) or fully valid endpoint with dummy region assumed (e.g
 // for URLs to emulated services).
 func ConfigFromURL(awsURL *url.URL) (*aws.Config, error) {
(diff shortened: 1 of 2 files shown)


bboreham:
What is the failure mode now if you forget to supply credentials via parameter or environment?

tomwilkie:
Assuming there are no credentials in the environment it'll still fail on startup, with something along the lines of "No credentials in chain".

If you run it on AWS through, there are always some credentials available - they just might not allow you to read or write from dynamodb.  In this case the individual reads & writes will fail.

tomwilkie:
ping @bboreham?

bboreham:
I was trying to get more feedback from users of this library.

bboreham:
Thanks @tomwilkie !

PR #113, opened 2018-07-27
Set $SRC_PATH as required by the build container

bboreham:
This was missed in the earlier change (#98) to use the image from build-tools.


gouthamve:
LGTM +1

PR #114, opened 2018-07-27
Expose jaeger metrics

gouthamve:
@tomwilkie @bboreham


leth:
Are there any good resources on what metrics this will expose?

tomwilkie:
Nices, looks useful.

gouthamve:
@leth These are all the metrics: https://github.com/jaegertracing/jaeger-client-go/blob/master/metrics.go#L22

screen shot 2018-07-30 at 10 21 05

PR #117, opened 2018-08-23
Update httpgrpc to match weaveworks/cortex#910

bboreham:
A couple of files were mistakenly updated inside the vendor directory in Cortex.
https://github.com/weaveworks/cortex/pull/910/commits/f807690176ec98508a03c5ba29cbd8740c309ebd
This PR aims to replicate those changes upstream to un-break dependency management.


tomwilkie:
LGTM (just need to add comments on the new methods)

PR #118, opened 2018-08-23
Expose the HTTP server from the server struct.

tomwilkie:
This allows users to inject HTTP requests and have the logged, instrumented and traced correctly.

See https://github.com/weaveworks/cortex/issues/949

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -66,9 +66,8 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 type Server struct {
 	cfg          Config
 	handler      *signals.Handler
-	httpListener net.Listener
 	grpcListener net.Listener
-	httpServer   *http.Server
+	httpListener net.Listener
 
 	HTTP       *mux.Router
 	HTTPServer *http.Server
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -70,9 +70,10 @@ type Server struct {
 	grpcListener net.Listener
 	httpServer   *http.Server
 
-	HTTP *mux.Router
-	GRPC *grpc.Server
-	Log  logging.Interface
+	HTTP       *mux.Router
+	HTTPServer *http.Server
```
bboreham: why do we want both `httpServer` and `HTTPServer` ?
Is the exposed one supposed to be mutable?
If not, a function would be better.

tomwilkie: > why do we want both httpServer and HTTPServer ?

Good point.

> Is the exposed one supposed to be mutable?

The HTTPServer isn't supposed to be mutable, no.  We just need to be able to call the (middleware wrapped) handler on it; we could just expose that instead?

bboreham:
note CI tests failed

tomwilkie:
I think its flake #120

tomwilkie:
Thanks!

PR #124, opened 2018-09-13
Don't trace http requests when using httpgrpc.

tomwilkie:
This was introduced in #105, and breaks query-frontend -> querier trace propagation in cortex - specifically, this inserts a span between the client and server span, causing them not to be associated with each other in the jaeger UI correctly.  @leth is there any background on why this was nessecary?

The way I see it, when using httpgrpc you should propagate your context using grpc (using https://github.com/grpc-ecosystem/grpc-opentracing/tree/master/go/otgrpc middleware on the grpc clients and servers).  And when using http, you should propagate your context using the nethttp library (https://github.com/opentracing-contrib/go-stdlib).  But you shouldn't mix them.  

You can simulate the effect of this PR by wrapping the handler passed to httpgrpc in the nethttp middleware if you need to.

### DIFF
--- a/httpgrpc/server/server.go
+++ b/httpgrpc/server/server.go
@@ -14,7 +14,6 @@ import (
 	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
 	"github.com/mwitkow/go-grpc-middleware"
 	"github.com/opentracing/opentracing-go"
-	"github.com/opentracing/opentracing-go/ext"
 	"github.com/sercand/kuberesolver"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


leth:
> is there any background on why this was nessecary?

We were not seeing traces from authfe into cortex, the "please sample this" decision was not being passed along. #105 fixed it, but perhaps the real problem is elsewhere in authfe.

tomwilkie:
I see; does authfe have the grpc opentracing middleware on the clients?  It should do if its using https://github.com/weaveworks/common/blob/master/httpgrpc/server/server.go#L130.   Are you putting the tracing context in the downstream http request context?

leth:
>does authfe have the grpc opentracing middleware on the clients? It should do if its using https://github.com/weaveworks/common/blob/master/httpgrpc/server/server.go#L130.

Yep it's using that.

>Are you putting the tracing context in the downstream http request context?

Perhaps not? I'm working on a unit test for authfe to isolate our problem 👍

tomwilkie:
Cool, let me know if you need a hand.

PR #126, opened 2018-09-14
Pass options to tracing middleware

### LINKED ISSUE #125: Can't pass in options to opentracing handler
So, the query endpoint AND the `/metrics` endpoint both end up under the operation: "HTTP GET" and this clutters up a traces a fair bit.

We could pass in something like this: https://godoc.org/github.com/opentracing-contrib/go-stdlib/nethttp#OperationNameFunc to differentiate both but because opentracing is a default middleware, there is no way to do that, see: https://github.com/weaveworks/common/blob/master/middleware/http_tracing.go#L17

Now, we can pass in the options to the tracer, but would you open to adding something like `OpentracingOptions` to `server.Config`?

gouthamve:
Fixes #125 

@tomwilkie @bboreham

### DIFF
--- a/middleware/http_tracing.go
+++ b/middleware/http_tracing.go
@@ -10,11 +10,13 @@ import (
 )
 
 // Tracer is a middleware which traces incoming requests.
-type Tracer struct{}
+type Tracer struct {
+	Options []nethttp.MWOption
+}
 
 // Wrap implements Interface
 func (t Tracer) Wrap(next http.Handler) http.Handler {
-	traceHandler := nethttp.Middleware(opentracing.GlobalTracer(), next)
+	traceHandler := nethttp.Middleware(opentracing.GlobalTracer(), next, t.Options...)
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		var maybeTracer http.Handler
 		// Don't try and trace websocket requests because nethttp.Middleware
--- a/server/server.go
+++ b/server/server.go
@@ -8,6 +8,8 @@ import (
 	_ "net/http/pprof" // anonymous import to get the pprof handler registered
 	"time"
 
+	"github.com/opentracing-contrib/go-stdlib/nethttp"
+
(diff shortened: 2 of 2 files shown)


server/server.go
```diff
@@ -42,6 +44,7 @@ type Config struct {
 	GRPCMiddleware       []grpc.UnaryServerInterceptor
 	GRPCStreamMiddleware []grpc.StreamServerInterceptor
 	HTTPMiddleware       []middleware.Interface
+	OpentracingOptions   []nethttp.MWOption
```
bboreham: These look more like nethttp middleware options than Opentracing options

server/server.go
```diff
@@ -139,7 +142,9 @@ func New(cfg Config) (*Server, error) {
 		RegisterInstrumentation(router)
 	}
 	httpMiddleware := []middleware.Interface{
-		middleware.Tracer{},
+		middleware.Tracer{
+			Options: cfg.OpentracingOptions,
```
bboreham: Maybe the caller should pass in a middleware.Interface to act as Tracer, similar to how Log is done?

gouthamve: What do you mean? Could you explain? The `Log` is being passed a `logger`...

gouthamve:
@tomwilkie That works :)

bboreham:
The code changes no longer match the PR title or description.  Please update something.

gouthamve:
@bboreham Sorry, updated the title and description.

middleware/http_tracing.go
```diff
@@ -10,11 +11,22 @@ import (
 )
 
 // Tracer is a middleware which traces incoming requests.
-type Tracer struct{}
+type Tracer struct {
```
leth: I'm wondering if we should rename `Tracer`, but can't really come up with a compelling enough alternative:

- RequestTracer
- OpentracingMiddleware
- TracingRequestHandler

🤢

PR #127, opened 2018-09-19
 Update more golang/protobuf to gogo/protobuf

bboreham:
It seems to be necessary to decode errors as `github.com/gogo/protobuf/types.Any` rather than `github.com/golang/protobuf/ptypes/any`, following #117, otherwise we get failures to decode errors.

It is possible this relates more to versions of the libs than the specific code - there are a number of issues in the gogo repos around `Any`, and I had to add some overrides in a downstream repo to get the latest:

```
[[override]]
  name = "github.com/gogo/googleapis"
  version = "v1.1.0"

# Need this version to get past https://github.com/gogo/status/issues/4 
[[override]]
  name = "github.com/gogo/status"
  version = "v1.0.3"
```

### DIFF
--- a/httpgrpc/httpgrpc.go
+++ b/httpgrpc/httpgrpc.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 
 	spb "github.com/gogo/googleapis/google/rpc"
-	ptypes "github.com/gogo/protobuf/types"
+	"github.com/gogo/protobuf/types"
 	"github.com/gogo/status"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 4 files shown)


httpgrpc/httpgrpc.go
```diff
@@ -3,11 +3,10 @@ package httpgrpc
 import (
 	"fmt"
 
-	"github.com/golang/protobuf/ptypes"
-	"github.com/golang/protobuf/ptypes/any"
+	spb "github.com/gogo/googleapis/google/rpc"
+	ptypes "github.com/gogo/protobuf/types"
```
tomwilkie: Whats the reason for the import rename?

bboreham: So I didn't have to edit the lines that said `ptypes.MarshalAny` and `ptypes.UnmarshalAny`

tomwilkie: Oh okay; I'd personally just change these lines, but not a big deal.

tomwilkie:
I think if you put the `[[override]]` in as a `[[constraint]]` in this repo, dep should "inherit" it in any downstream usage.  I thought that was the whole point of dep...

bboreham:
Funnily in this repo `dep` manages to pick the right versions, but you're right I could add a constraint here.

httpgrpc/httpgrpc.go
bboreham: changed

PR #128, opened 2018-09-19
Fix dropped namespace in url parsing

leth:
Causes problems if you try and connect from one namespace to another.

### DIFF
--- a/httpgrpc/server/server.go
+++ b/httpgrpc/server/server.go
@@ -91,13 +91,17 @@ func ParseURL(unparsed string) (string, []grpc.DialOption, error) {
 		if err != nil {
 			return "", nil, err
 		}
-		parts := strings.SplitN(host, ".", 2)
-		service, namespace := parts[0], "default"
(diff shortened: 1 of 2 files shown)


tomwilkie:
How so?  These addresses are interpreted by kuberesolver as a k8s service name here: https://github.com/sercand/kuberesolver/blob/v1.0.0/balancer.go#L46

By putting a dot in there, you're telling kuberesolver to pick a port name I think... which does leave the question how the port number syntax ever worked.

leth:
> How so? 

Connecting to `foo.bar.svc.cluster.local` results in a connection to `foo` + the current namespace, this works if your current namespace is also `bar`, but not otherwise.

> By putting a dot in there, you're telling kuberesolver to pick a port name I think... which does leave the question how the port number syntax ever worked.

I think you mixed up `.` and `:`?

bboreham:
Empirically, we omit `.svc.cluster.local` when using this feature and connect to `querier.cortex` or whatever.

PR #131, opened 2018-11-19
Update to latest kuberesolver

gouthamve:
@tomwilkie @bboreham

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -10,19 +10,20 @@
   revision = "e75332964ef517daa070d7c38a9466a0d687e0a5"
 
 [[projects]]
-  digest = "1:cd2a14ee6212fc0353b02dc489c32fa666882de0e57bd36d6d56e98d9f651974"
+  digest = "1:1f6e8e5dc5378c426699ea8788cecc07cf99a4d72402341da0638f6607d364f7"
   name = "github.com/aws/aws-sdk-go"
   packages = [
     "aws",
     "aws/awserr",
     "aws/credentials",
     "aws/endpoints",
+    "internal/ini",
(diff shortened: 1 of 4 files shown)


gouthamve:
Nope, but it uses the new gRPC API for loadblancing. It's just the updated version of the previous library.

bboreham:
Review might go faster if you say what is going on in the 80-odd lines of changes.

gouthamve:
Previously, we needed to pass in a resolver as an option. And the resolver used to do both resolution and loadbalancing, afaik.

Now, the API changed to "registering" resolvers. i.e, mapping scheme to resolvers, for example `kubernetes://` --> kuberesolver, `dns://` --> dns resolver, etc. And then specifying the load-balancing policy via the resolution mechanism or manually or via an `grpclb`, etc.

See: https://github.com/grpc/proposal/blob/master/L9-go-resolver-balancer-API.md

In this PR, I modified our client creation to use the new API through kuberesolver v2. It was fairly straightforward, needed to remove the additional resolver dial option and needed to specify the loadbalancing.

tomwilkie:
I was also hoping that kuberesolver v2 would fix https://github.com/sercand/kuberesolver/issues/4, but it doesn't appear to.

bboreham:
> for example kubernetes:// --> kuberesolver, dns:// --> dns resolver, etc. 

According to https://github.com/grpc/grpc-go/blob/v1.19.0/clientconn.go#L125 those should have three slashes.  Did you test this code?

gouthamve:
Oh right, we ended up using `dns:///` which worked. Though let me update the code to make it `:///`

PR #132, opened 2018-11-30
Allow server.Config to be unmarshaled from YAML.

tomwilkie:
This is so we can load flags from a YAML file in Cortex.

### DIFF
--- a/logging/level.go
+++ b/logging/level.go
@@ -40,7 +40,16 @@ func (l *Level) String() string {
 	return l.s
 }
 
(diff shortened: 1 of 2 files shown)


bboreham:
Is the second commit supposed to be here ("Allow overriding max message size in the server")?  Seems unrelated to the PR description.

tomwilkie:
Yeah, my bad.  Moved to #146.

tomwilkie:
@bboreham I can haz merge pls? :-)

PR #133, opened 2018-12-01
Use vanilla go test ./...

tomwilkie:
- it ignores vendor now
- it still tries to build packages even if they don't have tests.

I'm doing this as I noticed if you start a new project and use weaveworks/common, you run into:

```
# github.com/weaveworks/common/vendor/github.com/uber/jaeger-lib/metrics/prometheus
vendor/github.com/uber/jaeger-lib/metrics/prometheus/factory.go:143:3: cannot use hv.WithLabelValues(f.tagsAsLabelValues(labelNames, tags)...) (type prometheus.Observer) as type prometheus.Histogram in field value:
	prometheus.Observer does not implement prometheus.Histogram (missing Collect method)
```

Which we've have to fix a couple of time already. 

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -34,14 +34,6 @@
   revision = "3a771d992973f24aa725d07868b467d1ddfceafb"
 
 [[projects]]
-  branch = "master"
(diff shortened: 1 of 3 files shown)


tomwilkie:
(test on 2964cdb did fail, don't know when github not showing it: https://circleci.com/gh/tomwilkie/weaveworks-common/96)

PR #145, opened 2018-12-31
Use prometheus.ObserverVec so instrument package can be used with cur…

tomwilkie:
…ried labels.

### DIFF
--- a/instrument/instrument.go
+++ b/instrument/instrument.go
@@ -26,7 +26,7 @@ type Collector interface {
 
 // HistogramCollector collects the duration of a request
 type HistogramCollector struct {
-	metric *prometheus.HistogramVec
+	metric prometheus.ObserverVec
 }
 
 // HistogramCollectorBuckets define the buckets when passing the metric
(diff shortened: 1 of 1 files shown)


bboreham:
Rather than making a breaking change to a deprecated interface, could the code using this interface be updated to use the recommended interface?

tomwilkie:
> Rather than making a breaking change to a deprecated interface

How is it a breaking change?  IIUC `HistogramVec` implements `ObserverVec`, which guarantees all previous callers can call the new function.

The only way I can think is would break is if downstream code saved this function into an explicitly typed variable (of the old type).   I can't find anyone that does that (on Github), and it sounds super niche.

bboreham:
Having thought about this some more (and done the update in https://github.com/cortexproject/cortex/pull/1153), this change leaves `HistogramCollector` as an unhappy mashup - it can be created via `NewHistogramCollectorFromOpts` whereby it will definitely be histograms, or via `NewHistogramCollector` in which case it might be some other vector of observers.

I suspect that `Collector` and `ObserverVec` are both doing similar indirections, though the interfaces are sufficiently far apart that it isn't obvious how to remove one.  That leaves `type ObserverCollector` as a way forward.  And you can implement that on the calling side; you don't need to put it in the common library.

PR #148, opened 2019-01-08
Dont flag context.Canceled as an error in tracing

bboreham:
Cancellations are always caused by something else, so having them show up as errors in the tracing UI is distracting.


tomwilkie:
LGTM!

PR #151, opened 2019-03-17
Skip logging context cancelled warning msgs

nrjpoddar:
Context cancellations are controlled by clients and logging this message can
cause more more confusion as mentioned here:
https://github.com/cortexproject/cortex/issues/1279

Signed-off-by: Neeraj Poddar <neeraj@aspenmesh.io>

PR #159, opened 2019-06-17
Updated handling of gRPC Context Cancelled

nrjpoddar:
Signed-off-by: Neeraj Poddar <neeraj@aspenmesh.io>

### DIFF
--- a/middleware/grpc_logging.go
+++ b/middleware/grpc_logging.go
@@ -5,6 +5,8 @@ import (
 
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/weaveworks/common/logging"
 	"github.com/weaveworks/common/user"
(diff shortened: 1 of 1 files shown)


nrjpoddar:
This PR https://github.com/weaveworks/common/pull/151 was supposed to fix it but I noticed that the logging was still happening as the error checking was incorrect.

nrjpoddar:
@bboreham do you mind taking a look?

bboreham:
The code looks fine, but given the previous one also looked fine, can we get any more assurance this time?
How did you find out what to do, both times?  Can we point to a reference?
Is it feasible to write a test?

nrjpoddar:
Haha, totally agree and a valid point.

How did you find out what to do, both times? Can we point to a reference? -> First time I didn't understand correctly what error type was being returned here for context cancellations. After we upgraded ingester in our environment which had my fixes I was still seeing this error:

`ingester level=warn ts=2019-06-21T15:54:23.748137807Z caller=grpc_logging.go:54 method=/cortex.Ingester/QueryStream duration=59.418013ms err="rpc error: code = Canceled desc = context canceled" msg="gRPC\n"`

From this I derived that the error was a gRPC error type.

Unit test for this might be a bit tricky but I think I can write it with sufficient dependency injection.

bboreham:
Having extended the tests to cover cancellation, I believe this PR was incorrect - we need to cover both `context.Canceled` and `grpc/codes.Canceled`.  I created #180 to replace.

PR #161, opened 2019-07-12
Update prometheus/client_golang and prometheus/procfs

codesome:
With this, `prometheus.Handler()` became `promhttp.Handler()`.

PR #162, opened 2019-07-13
update vendored prom client and use module vendoring

jtlisi:
This PR updates the prometheus client_golang library and also updates the repo to use go modules.


jtlisi:
#161 makes this unnecessary

PR #164, opened 2019-08-16
use https if provided as scheme

thorfour:
If the scheme was set to https, respect https otherwise default to http as always.

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -45,6 +45,9 @@ func ConfigFromURL(awsURL *url.URL) (*aws.Config, error) {
 	}
 
 	if strings.Contains(awsURL.Host, ".") {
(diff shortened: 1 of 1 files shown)


bboreham:
Do you know what that code is intended to achieve?  Including "dummy" ?

If so, I'd appreciate a comment.

Your change LGTM from my current position of ignorance.

thorfour:
The AWS SDK requires a region to be set, but regardless of region set it will be ignored and utilize the region the endpoint refers to; hence "dummy", I was only copying the fake region that was already utilized. 

DigitalOcean spaces doesn't not support making `http` requests, so when using an `https` endpoint it gets overridden by this previous code and ends up failing. This was an attempt to allow the sane default of http to continue but to support https

PR #166, opened 2019-09-16
Allow region to be read from environment variable

### LINKED ISSUE #165: Setting AWS_REGION has no effect with custom endpoint
Please see the original issue: https://github.com/grafana/loki/issues/759
The problem seems to be located with the logic at: 
 > https://github.com/weaveworks/common/blob/master/aws/config.go#L48

adityacs:
Allow region to be read from environment variable.

For using custom s3 endpoint, some providers expects valid endpoint with valid `region`. This change should allow to use region from environment variable. If not defined it will fallback to `dummy`

Fixes #165

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


adityacs:
@bboreham Kindly review this.

aws/config.go
```diff
@@ -45,10 +46,14 @@ func ConfigFromURL(awsURL *url.URL) (*aws.Config, error) {
 	}
 
 	if strings.Contains(awsURL.Host, ".") {
+		region := "dummy"
+		if os.Getenv("AWS_REGION") != "" {
+			region = os.Getenv("AWS_REGION")
```
bboreham: I raised an eyebrow at the two identical calls - what about:

		region := os.Getenv("AWS_REGION")
		if  region == "" {
			region = "dummy"
		}

adityacs: Perfect!

adityacs:
@bboreham have made the changes. Kindly review.

PR #170, opened 2019-11-15
Limit the number of connections to server

jtlisi:
We ran into a situation where it would be good to limit the total number of connections. This PR adds an optional configuration to limit the number of simultaneous connections to the server. The package used to enable this is the [netutil LimitListener](https://godoc.org/golang.org/x/net/netutil#LimitListener).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -120,12 +120,12 @@
   version = "v1.6.2"
 
 [[projects]]
-  digest = "1:70c1f58286e9c70b080d2a41b1ec187d7fd7f7cc09d4c1f64b2ca205b7d9bbc9"
+  digest = "1:57710d97c66b36f74586f98ba18d2a10f40d85d9c090be66e67df82993f57e3c"
   name = "github.com/grpc-ecosystem/go-grpc-middleware"
   packages = ["."]
   pruneopts = "N"
-  revision = "c250d6563d4d4c20252cd865923440e829844f4e"
-  version = "v1.0.0"
+  revision = "dd15ed025b6054e5253963e355991f3070d4e593"
+  version = "v1.1.0"
 
 [[projects]]
   digest = "1:4643b94ee9e1e291b3f35558faddf6882cb7c2c759a416c0427bc13c2b992a24"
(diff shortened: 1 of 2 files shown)


tomwilkie:
@bboreham mind taking a look?  LGTM.

Gopkg.toml
```diff
@@ -25,3 +25,7 @@
 [[constraint]]
   name = "github.com/gogo/status"
   version = ">=v1.0.3"
+
+[[constraint]]
```
bboreham: Please add a comment explaining the need for this constraint, and the conditions under which it can be removed.

Gopkg.lock
```diff
@@ -19,7 +19,7 @@
     "aws/endpoints",
     "internal/ini",
     "internal/sdkio",
-    "internal/shareddefaults",
+    "internal/shareddefaults"
```
bboreham: This seems like a spurious and unnecessary change.  Do you know where it came from?

server/server.go
```diff
@@ -60,8 +63,10 @@ type Config struct {
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.HTTPListenAddress, "server.http-listen-address", "", "HTTP server listen address.")
 	f.IntVar(&cfg.HTTPListenPort, "server.http-listen-port", 80, "HTTP server listen port.")
+	f.IntVar(&cfg.HTTPConnLimit, "server.http-conn-limit", 0, "Maximum number of simultaneously http connections, <=0 to disable")
```
bboreham: s/simultaneously/simultaneous/

Gopkg.lock
jtlisi: This was done automatically up running `dep ensure`

server/server.go
jtlisi: 👍

jtlisi:
The constraint was added automatically by dep for some reason. However, it does not seem to be required so I removed it and re-ran `dep ensure`.

@bboreham This should be good for a second look

PR #174, opened 2020-01-13
add flags to configure keep alive configs

jtlisi:
We are running into issues with with load balancing the distributors in cortex. We are using DNS resolution with the httpgrpc client. However, the default resolution frequency of the client is 30 minutes. Since we don't want to roll a custom resolver, the other recommended approach is to set the `max_connection_age` of the grpc server (https://github.com/grpc/grpc/issues/12295).

This PR entails the following:
* Flags to set the GRPC Server Keepalive configs that will be passed as options upon creation
* Flags default to the default values of the `grpc-go` package
* Because we don't roll the config struct itself, there are no `yaml` tags, so it won't be able to be read from a config file

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -3,6 +3,7 @@ package server
 import (
 	"flag"
 	"fmt"
+	math "math"
 	"net"
 	"net/http"
 	_ "net/http/pprof" // anonymous import to get the pprof handler registered
@@ -16,6 +17,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/keepalive"
 
 	"github.com/weaveworks/common/httpgrpc"
 	httpgrpc_server "github.com/weaveworks/common/httpgrpc/server"
@@ -46,9 +48,10 @@ type Config struct {
 	GRPCStreamMiddleware []grpc.StreamServerInterceptor `yaml:"-"`
 	HTTPMiddleware       []middleware.Interface         `yaml:"-"`
 
-	GPRCServerMaxRecvMsgSize       int  `yaml:"grpc_server_max_recv_msg_size"`
-	GRPCServerMaxSendMsgSize       int  `yaml:"grpc_server_max_send_msg_size"`
-	GPRCServerMaxConcurrentStreams uint `yaml:"grpc_server_max_concurrent_streams"`
+	GPRCServerMaxRecvMsgSize       int                        `yaml:"grpc_server_max_recv_msg_size"`
+	GRPCServerMaxSendMsgSize       int                        `yaml:"grpc_server_max_send_msg_size"`
+	GPRCServerMaxConcurrentStreams uint                       `yaml:"grpc_server_max_concurrent_streams"`
+	GRPCServerKeepAliveConfigs     keepalive.ServerParameters `yaml:"-"`
 
 	LogLevel logging.Level     `yaml:"log_level"`
 	Log      logging.Interface `yaml:"-"`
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -46,9 +48,10 @@ type Config struct {
 	GRPCStreamMiddleware []grpc.StreamServerInterceptor `yaml:"-"`
 	HTTPMiddleware       []middleware.Interface         `yaml:"-"`
 
-	GPRCServerMaxRecvMsgSize       int  `yaml:"grpc_server_max_recv_msg_size"`
-	GRPCServerMaxSendMsgSize       int  `yaml:"grpc_server_max_send_msg_size"`
-	GPRCServerMaxConcurrentStreams uint `yaml:"grpc_server_max_concurrent_streams"`
+	GPRCServerMaxRecvMsgSize       int                        `yaml:"grpc_server_max_recv_msg_size"`
+	GRPCServerMaxSendMsgSize       int                        `yaml:"grpc_server_max_send_msg_size"`
+	GPRCServerMaxConcurrentStreams uint                       `yaml:"grpc_server_max_concurrent_streams"`
+	GRPCServerKeepAliveConfigs     keepalive.ServerParameters `yaml:"-"`
```
pracucci: Ideally we should allow to configure all parameters from YAML as well. What do you think to move here the 5 params we configure in `GRPCServerKeepAliveConfigs` and then when we call `grpc.KeepaliveParams()` we do create a `keepalive.ServerParameters{}` on the fly using the 5 params set here?

jtlisi: That sounds good to me since we can't add struct tags to the underlying struct.

server/server.go
```diff
@@ -46,16 +48,23 @@ type Config struct {
 	GRPCStreamMiddleware []grpc.StreamServerInterceptor `yaml:"-"`
 	HTTPMiddleware       []middleware.Interface         `yaml:"-"`
 
-	GPRCServerMaxRecvMsgSize       int  `yaml:"grpc_server_max_recv_msg_size"`
-	GRPCServerMaxSendMsgSize       int  `yaml:"grpc_server_max_send_msg_size"`
-	GPRCServerMaxConcurrentStreams uint `yaml:"grpc_server_max_concurrent_streams"`
+	GPRCServerMaxRecvMsgSize        int           `yaml:"grpc_server_max_recv_msg_size"`
+	GRPCServerMaxSendMsgSize        int           `yaml:"grpc_server_max_send_msg_size"`
+	GPRCServerMaxConcurrentStreams  uint          `yaml:"grpc_server_max_concurrent_streams"`
+	GRPCServerMaxConnectionIdle     time.Duration `yaml:"grpc_max_connection_idle"`
```
pracucci: To keep consistency with the existing yaml options, maybe it's better if prefix the YAML fields with `grpc_server_`  instead of just `grpc_`.

server/server.go
```diff
@@ -70,6 +79,11 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
 	f.UintVar(&cfg.GPRCServerMaxConcurrentStreams, "server.grpc-max-concurrent-streams", 100, "Limit on the number of concurrent streams for gRPC calls (0 = unlimited)")
+	flag.DurationVar(&cfg.GRPCServerMaxConnectionIdle, "server.grpc.keepalive.max-connection-idle", infinty, "The duration after which an idle connection should be closed. Default: infinity")
+	flag.DurationVar(&cfg.GRPCServerMaxConnectionAge, "server.grpc.keepalive.max-connection-age", infinty, "The duration for the maximum amount of time a connection may exist before it will be closed. Default: infinity")
+	flag.DurationVar(&cfg.GRPCServerMaxConnectionAgeGrace, "server.grpc.keepalive.max-connection-age-grace", infinty, "An additive period after max-connection-age after which the connection will be forcibly closed. Default: infinity")
+	flag.DurationVar(&cfg.GRPCServerTime, "server.grpc.keepalive.time", time.Hour*2, "Duration after which a keepalive probe is sent in case of no activity over the connection., Default: 2h")
+	flag.DurationVar(&cfg.GRPCServerTimeout, "server.grpc.keepalive.timeout", time.Second*20, "Duration after which an idle connection should be closed, Default: 20s")
```
bboreham: I was a bit confused by this description; the gRPC comment "After having pinged for keepalive check" makes it clearer.

pracucci:
@bboreham Thanks for your review here. Given Jacob has fixed the nit, would you see feasible to merge it so that we would update the dependency on the Cortex side, please?

PR #175, opened 2020-01-15
Fix kuberesolver URLs to have three slashes

bboreham:
According to https://github.com/grpc/grpc-go/blob/v1.19.0/clientconn.go#L125 the URLs to pass to gRPC should have three slashes. 

Empirically this works better in my environment.

### DIFF
--- a/httpgrpc/server/server_test.go
+++ b/httpgrpc/server/server_test.go
@@ -95,9 +95,10 @@ func TestParseURL(t *testing.T) {
 		err      error
 	}{
 		{"direct://foo", "foo", nil},
-		{"kubernetes://foo:123", "kubernetes://foo:123", nil},
-		{"querier.cortex:995", "kubernetes://querier.cortex:995", nil},
-		{"foo.bar.svc.local:995", "kubernetes://foo.bar.svc.local:995", nil},
+		{"kubernetes://foo:123", "kubernetes:///foo:123", nil},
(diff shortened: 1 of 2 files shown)


gouthamve:
Haha, sorry I missed this when I opened: https://github.com/weaveworks/common/pull/176/files Closing that.

httpgrpc/server/server_test.go
```diff
@@ -95,9 +95,10 @@ func TestParseURL(t *testing.T) {
 		err      error
 	}{
 		{"direct://foo", "foo", nil},
-		{"kubernetes://foo:123", "kubernetes://foo:123", nil},
-		{"querier.cortex:995", "kubernetes://querier.cortex:995", nil},
-		{"foo.bar.svc.local:995", "kubernetes://foo.bar.svc.local:995", nil},
+		{"kubernetes://foo:123", "kubernetes:///foo:123", nil},
+		{"querier.cortex:995", "kubernetes:///querier.cortex:995", nil},
+		{"foo.bar.svc.local:995", "kubernetes:///foo.bar.svc.local:995", nil},
+		{"kubernetes:///foo:123", "kubernetes:///foo:123", nil},
```
gouthamve: Can we also add `dns:///foo:123` to the list so that we are not k8s specific, but also have another resolver in the list.

bboreham: Done (I copied the line from your PR).

PR #180, opened 2020-02-18
Improve handling of canceled calls in logging, metrics and tracing.

bboreham:
Replacement for #159, including testing.

The visible improvement is that gRPC streaming functions no longer cause a log message when canceled by the caller.

Canceled functions of all kinds no longer show as an error in tracing.

Also note that canceled gRPC functions will now show up in metrics under the label `cancel` rather than `error`.

There are a lot of diffs from the re-generation of a protobuf file, which I separated to its own commit.


grpc/cancel.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"context"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// IsCanceled checks whether an error comes from an operation being canceled
+func IsCanceled(err error) bool {
+	if err == context.Canceled {
```
thorfour: Maybe use `errors.Is`?

bboreham: That would force consumers to upgrade to Go 1.13, I think.
Which is not a bad idea, but more work than I want to complete before merging this.
Do you want to file it as an issue/enhancement?

thorfour: That's a good consideration. I can file it as an enhancement

PR #182, opened 2020-02-24
BREAKING CHANGE: Return error in configuring jaeger

### LINKED ISSUE #181: Jaeger agent setup error
When configuring jaeger tracing. It would be good to return the error to caller rather than exiting it. The client can decide if they want to exit or still continue.

FYR: https://github.com/grafana/loki/issues/1397

adityacs:
Fixes #181 

Also, we should add a watcher so that when jaeger is up, tracing can be initialized. Should we leave this logic to client or can we implement it here?

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -1,24 +1,26 @@
 package tracing
 
 import (
-	"fmt"
 	"io"
-	"io/ioutil"
-	"os"
 
+	"github.com/pkg/errors"
 	jaegercfg "github.com/uber/jaeger-client-go/config"
 	jaegerprom "github.com/uber/jaeger-lib/metrics/prometheus"
 )
 
+// ErrInvalidConfiguration is an error to notify client to provide valid trace report agent or config server
+var (
+	ErrBlankTraceConfiguration = errors.New("no trace report agent or config server specified")
+)
+
 // installJaeger registers Jaeger as the OpenTracing implementation.
-func installJaeger(serviceName string, cfg *jaegercfg.Configuration) io.Closer {
+func installJaeger(serviceName string, cfg *jaegercfg.Configuration) (io.Closer, error) {
 	metricsFactory := jaegerprom.New()
 	closer, err := cfg.InitGlobalTracer(serviceName, jaegercfg.Metrics(metricsFactory))
 	if err != nil {
-		fmt.Printf("Could not initialize jaeger tracer: %s\n", err.Error())
-		os.Exit(1)
+		return nil, errors.Wrap(err, "could not initialize jaeger tracer")
 	}
-	return closer
+	return closer, nil
 }
 
 // NewFromEnv is a convenience function to allow tracing configuration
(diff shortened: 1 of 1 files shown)


bboreham:
Whilst I agree in principle, I think it's important to be a bit louder about breaking changes in a widely-used library.

How about changing the function name?

adityacs:
@bboreham Agree
Can we can add a config where the client can choose to use the current implementation where it exits on jaeger not being setup 
OR
Choose to watch for jaeger agent and initialize when jaeger is up ?

bboreham:
> Can we can add a config where the client can choose to use the current implementation where it exits on jaeger not being setup

Too complicated.

> OR
> Choose to watch for jaeger agent and initialize when jaeger is up ?

It won't fail just because the Jaeger agent isn't there.

adityacs:
If the configuration parameters are wrong, then we should leave it to user to take care of correct configuration parameters. This is not user friendly feature. There would be no way we can tell user configuration parameters are wrong.

Hmm, this is little tricky to handle. 

Can we add retry and timeout? Not sure, just looking for ideas

bboreham:
My idea is: change the name of the function, so that people can discover that the behaviour changed.

bboreham:
Thinking again, most people will be using the return value, so the change to two return values will break their build and they have to look at what changed.

tracing/tracing.go
```diff
@@ -11,14 +11,14 @@ import (
...
 
 // installJaeger registers Jaeger as the OpenTracing implementation.
-func installJaeger(serviceName string, cfg *jaegercfg.Configuration) io.Closer {
+func installJaeger(serviceName string, cfg *jaegercfg.Configuration) (io.Closer, error) {
 	metricsFactory := jaegerprom.New()
 	closer, err := cfg.InitGlobalTracer(serviceName, jaegercfg.Metrics(metricsFactory))
 	if err != nil {
 		fmt.Printf("Could not initialize jaeger tracer: %s\n", err.Error())
-		os.Exit(1)
+		return nil, err
```
bboreham: Return the error, wrapped, rather than logging it here.

tracing/tracing.go
```diff
@@ -36,7 +36,7 @@ func NewFromEnv(serviceName string) io.Closer {
 
 	if cfg.Sampler.SamplingServerURL == "" && cfg.Reporter.LocalAgentHostPort == "" {
 		fmt.Printf("Jaeger tracer disabled: No trace report agent or config server specified\n")
-		return ioutil.NopCloser(nil)
+		return ioutil.NopCloser(nil), nil
```
bboreham: Feels like this should be an error return for consistency, perhaps to an exported variable so the caller can check which kind of error.

tracing/tracing.go
adityacs: Sure

tracing/tracing.go
adityacs: Will change

adityacs:
@bboreham Kindly validate the new changes

adityacs:
@bboreham Kindly requesting you to validate the new changes

tracing/tracing.go
```diff
@@ -1,24 +1,26 @@
...
-	"os"
 
+	"github.com/pkg/errors"
 	jaegercfg "github.com/uber/jaeger-client-go/config"
 	jaegerprom "github.com/uber/jaeger-lib/metrics/prometheus"
 )
 
+// ErrInvalidConfiguration is an error to notify client to provide valid trace report agent or config server
+var (
+	ErrInvalidTraceConfiguration = errors.New("no trace report agent or config server specified")
```
bboreham: I don't see this as "Invalid"; it was previously a legitemate way to disable tracing.
How about "Blank"?

adityacs: Yeah `Blank` makes sense. Have made the change.

PR #183, opened 2020-03-10
Fixed GRPCServerLog log level on gRPC canceled error

pracucci:
While working on https://github.com/cortexproject/cortex/pull/2242 I've realized that the gRPC server interceptor used to log errors does log with `debug` level if the error is `context.Canceled` but not if it's a gRPC error wrapping the canceled one.

This PR should fix it (I've manually tested it in Cortex against PR 2242).

### DIFF
--- a/middleware/grpc_logging.go
+++ b/middleware/grpc_logging.go
@@ -3,8 +3,10 @@ package middleware
 import (
 	"time"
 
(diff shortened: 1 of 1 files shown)


pracucci:
_Closing cause I'm an idiot and I haven't checked `master` (already fixed there)._

PR #184, opened 2020-04-01
add log format option

seankhliao:
Add option to specify log output format to be json

not sure if unknown formats should be rejected? I don't see the option of returning an error

Related grafana/loki#1885

### DIFF
--- a/logging/gokit.go
+++ b/logging/gokit.go
@@ -8,15 +8,23 @@ import (
 	"github.com/go-kit/kit/log/level"
 )
 
-// NewGoKit creates a new Interface backed by a GoKit logger
-func NewGoKit(l Level, format string) Interface {
-	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
-	if format == "json" {
+// NewGoKitFormat creates a new Interface backed by a GoKit logger
+// format can be "json" or defaults to logfmt
+func NewGoKitFormat(l Level, f Format) Interface {
+	var logger log.Logger
+	if f.s == "json" {
 		logger = log.NewJSONLogger(log.NewSyncWriter(os.Stderr))
+	} else {
+		logger = log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
 	}
 	logger = level.NewFilter(logger, l.Gokit)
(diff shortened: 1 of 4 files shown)


logging/gokit.go
```diff
@@ -9,8 +9,11 @@ import (
 )
 
 // NewGoKit creates a new Interface backed by a GoKit logger
-func NewGoKit(l Level) Interface {
+func NewGoKit(l Level, format string) Interface {
 	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
+	if format == "json" {
+		logger = log.NewJSONLogger(log.NewSyncWriter(os.Stderr))
```
bboreham, line 15: Seems kinda wasteful to create and immediately discard the other one; suggest `if/else`.

server/server.go
```diff
@@ -60,8 +60,9 @@ type Config struct {
 	GRPCServerTime                  time.Duration `yaml:"grpc_server_keepalive_time"`
 	GRPCServerTimeout               time.Duration `yaml:"grpc_server_keepalive_timeout"`
 
-	LogLevel logging.Level     `yaml:"log_level"`
-	Log      logging.Interface `yaml:"-"`
+	LogFormat string            `yaml:"log_format"`
```
bboreham, line 63: for consistency this should have a line in `RegisterFlags`.

seankhliao:
I'd actually like to know how the breaking change could be avoided, was really hesitant about making this PR because of it

bboreham:
Add a new function with two parameters and have the old name call the new one.

seankhliao:
I went with adding a new function as suggested and changed format to use the same setup as Level for config/flags

bboreham:
In case you didn't notice, it failed with a lint "error":
```
logging/logrus.go:9:1: comment on exported function NewLogrusFormat should be of the form "NewLogrusFormat ..."
```

Would also consider a PR changing to golangci-lint.

seankhliao:
Should be fixed now

gouthamve:
@bboreham Could you take another look here?

GiedriusS:
Any updates on this? Could we move it forward somehow? @bboreham :eye:

PR #185, opened 2020-04-09
Add TLS support to HTTP Server

annanay25:
Part of https://github.com/cortexproject/cortex/issues/2350

#### Changes
- Adds six new config flags which will accept path to the Cert, Key and ClientCA, for HTTP and GRPC servers.
  - `server.http-cert-path`, `server.http-key-path`, `server.http-ca-path`
  - `server.grpc-cert-path`, `server.grpc-key-path`, `server.grpc-ca-path` 
- Either both server and client auth will be enabled, or neither. 
- Added tests, and scripts to generate certs at runtime.
- Changes in dependencies are because of the following addition:
```
github.com/prometheus/node_exporter v1.0.0-rc.0.0.20200417100735-fa4edd700ebc
```
- `ConfigToTLSConfig`, which is used from the prometheus `node_exporter` library and generates a `*tls.Config` struct from the config options, is awaiting security review.
- The following breaking change has been made to the prometheus `client_golang` library:
```
github.com/prometheus/client_golang v1.4.1
```

/cc @bboreham


bboreham:
> Q: Should dummy certs for testing be included in this PR, or a script to generate certs at test runtime?

I don't know; what are the trade-offs?  What did someone else do when faced with this question?

annanay25:
Adding certs to the repo for testing purposes seems like the most common solution, the only trade-off being that the test will fail once the certs expire.

halcyondude:
I would like to confirm that both variants (enabled, not enabled) of this functionality will be tested in CI.

**Scenario**

Now that cortex is 1.0+ we plan to begin using it in production.  We have been rolling out a service mesh across our infra, for a few reasons - one of which is consistent mTLS across all k8s services which are on the mesh.  We expect this to be how we handle service <-> service encryption.  

While The motivation to bake this into the project is rational, I would hope that it doesn’t over time become mandatory as it adds another layer/place where certificates need to be managed.  The mesh we are using is able to integrate with external CA’s, and cert rotation/renewal is handled by cert-manager (using LetsEncrypt).

We are planning to run Cortex atop this existing infra, which in addition to providing mTLS, also enables observability, a layer of resiliency, and (I’m hoping) the ability to have blue-greens via traffic split for cortex service versions.

If I’m misunderstanding the use case here or intent apologies!  We are still learning about the project, code, and operational aspects of running cortex in our infra.

annanay25:
Thanks for your feedback, I completely agree with your reasoning @halcyondude. 

The PR has been designed to remain backward compatible. TLS remains optional, and can be setup via a service mesh as you have mentioned. If the relevant flags are not initialised, tls will not be initiated and the server will work exactly as before.

tomwilkie:
> Q: Should dummy certs for testing be included in this PR, or a script to generate certs at test runtime?

I think having certs (even dummy certs) checked in is undesirable; it makes it looks like a leak of secrets, even if it isn't.  Can we easily generate them in the unit test?

server/server.go
```diff
@@ -71,6 +78,9 @@ var infinty = time.Duration(math.MaxInt64)
 // RegisterFlags adds the flags required to config this to the given FlagSet
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.HTTPListenAddress, "server.http-listen-address", "", "HTTP server listen address.")
+	f.StringVar(&cfg.HTTPCertPath, "server.http-cert-path", "", "HTTP server cert path.")
+	f.StringVar(&cfg.HTTPKeyPath, "server.http-key-path", "", "HTTP server key path.")
+	f.StringVar(&cfg.HTTPKeyPath, "server.http-ca-path", "", "HTTP CA path.")
```
tomwilkie, line 83: This is also for the gRPC server, no?

server/server.go
```diff
@@ -128,6 +138,42 @@ func New(cfg Config) (*Server, error) {
...
+	if log == nil {
+		log = logging.NewLogrus(cfg.LogLevel)
+	}
+
+	// Setup TLS
+	var cert tls.Certificate
+	if cfg.HTTPCertPath != "" && cfg.HTTPKeyPath != "" {
+		cert, err = tls.LoadX509KeyPair(cfg.HTTPCertPath, cfg.HTTPKeyPath)
+		if err != nil {
+			log.Warnf("error loading cert %s or key %s, tls disabled", cfg.HTTPCertPath, cfg.HTTPKeyPath)
```
tomwilkie, line 153: We can't have the system start up insecure if loading the certs fail - we have to fail hard and return the error, maybe wrapping it with some context.  Same goes for below.

server/server.go
```diff
@@ -128,6 +138,42 @@ func New(cfg Config) (*Server, error) {
...
+	}
+
+	var tlsConfig *tls.Config
+	if len(cert.Certificate) > 0 && caCertPool != nil {
+		tlsConfig = &tls.Config{
+			Certificates: []tls.Certificate{cert},
+			ClientAuth:   tls.RequireAndVerifyClientCert,
+			ClientCAs:    caCertPool,
+		}
+	}
```
tomwilkie, line 175: Nit: I would put this all in a separate function which gives you a TLS config - might make it easier to test.

tomwilkie:
Can we reuse https://github.com/prometheus/node_exporter/blob/master/https/tls_config.go instead of building it ourself? This also has more options for client cert - which might be a good thing.

I think we'll need to be able to specify different TLS settings for gRPC and HTTP - for instance, not needing client cert verification on HTTP but needing it on gRPC sounds reasonable to me.

gouthamve:
I tried exposing the function upstream but I am not sure it will be done anytime soon https://github.com/prometheus/node_exporter/pull/1677#issuecomment-614483247

I don't think we should block this PR on that but we should add a TODO in the code and in an issue to move to the upstream function when its exposed. I also see that upstream is waiting for a security review, and I don't think we should block on that imo. The upstream code has already gone through robust code review here: https://github.com/prometheus/node_exporter/pull/1277 and from what I can see, by reusing that code, we're getting a free security review :)

server/server.go
```diff
@@ -15,9 +16,11 @@ import (
 	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	node_https "github.com/prometheus/node_exporter/https"
```
bboreham, line 19: Given the caution expressed in https://github.com/prometheus/node_exporter/pull/1677 I think there need to be some comments in the PR description and in the code noting that this code is awaiting security review.

server/server_test.go
```diff
@@ -287,3 +294,89 @@ func TestMiddlewareLogging(t *testing.T) {
...
+
+	fakeServer := FakeServer{}
+	RegisterFakeServerServer(server.GRPC, fakeServer)
+
+	go server.Run()
+	defer server.Shutdown()
+
+	clientCert, err := tls.LoadX509KeyPair("certs/client.crt", "certs/client.key")
+	if err != nil {
+		log.Warnf("error loading cert %s or key %s, tls disabled", "certs/client.crt", "certs/client.key")
```
bboreham, line 340: what does it mean for the test if this warning fires?

go.mod
```diff
@@ -4,49 +4,42 @@ go 1.13
...
 	github.com/opentracing-contrib/go-stdlib v0.0.0-20190519235532-cf7a6c988dc9
 	github.com/opentracing/opentracing-go v1.1.0
-	github.com/pkg/errors v0.8.1
+	github.com/pkg/errors v0.9.1
 	github.com/pmezard/go-difflib v1.0.0
-	github.com/prometheus/client_golang v0.9.2
-	github.com/prometheus/client_model v0.0.0-20190129233127-fd36f4220a90
-	github.com/prometheus/common v0.4.1
-	github.com/prometheus/procfs v0.0.3
+	github.com/prometheus/client_golang v1.4.1
```
bboreham, line 31: This is technically a breaking change - should be called out in the description.
Interestingly we were already on 1.x after #161, but maybe the conversion to go.mod reset the dependency.

go.mod
```diff
@@ -4,49 +4,42 @@ go 1.13
...
+	github.com/prometheus/common v0.9.1
+	github.com/prometheus/node_exporter v1.0.0-rc.0.0.20200417100735-fa4edd700ebc
 	github.com/sercand/kuberesolver v2.1.0+incompatible
 	github.com/sirupsen/logrus v1.4.2
 	github.com/stretchr/testify v1.4.0
 	github.com/uber/jaeger-client-go v2.15.0+incompatible
 	github.com/uber/jaeger-lib v1.5.1-0.20181102163054-1fc5c315e03c
 	github.com/weaveworks/promrus v1.2.0
-	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
```
bboreham, line 44: interesting that this one disappears, given the nature of the change

annanay25: This will now be vendored as part of the `node_exporter` repo.

server/server_test.go
annanay25: Changing this to `require.NoError()`

annanay25:
Updated PR description as well. Please let me know if this is sufficient to move forward. :)

server/certs/genCerts.sh
```diff
@@ -0,0 +1,23 @@
+# From https://github.com/joe-elliott/cert-exporter/blob/master/test/files/genCerts.sh
```
bboreham, line 1: Note this is under Apache 2 licence.
(also it's more future-proof to use a commit hash than `master` in the URL)

annanay25:
Thanks!

server/certs/genCerts.sh
annanay25: Ah, right. I could send a follow up PR to change this.

PR #186, opened 2020-04-14
Limit concurrent S3 connections per host

JensErat:
This commit limits the number of concurrent connections to encourage
connection reuse. `MaxIdleConnsPerHost` was already set to workaround
https://github.com/golang/go/issues/13801 ("issue which made Go consume
all ephemeral ports").

We observed a similar issue when starting up Loki (using weaveworks
`common/aws` throught Cortex) which resulted in thousands of very small
S3 requests until memcached was seeded. This failed because of ephemeral
port exhaustion: while the application in fact would have reused
connections, it already failed as it sent all of them concurrently until
most of the requests failed.

`MaxConnsPerHost` can be used to limit the overall number of parallel
connections.

An alternative solution would be to queue requests in the application, but the golang HTTP client already has a queue. Whether a maximum of 100 parallel requests is reasonable or too small might be subject of discussion. I'm not sure if there is a relevant use case of more than 100 parallel S3 streams per instance.

In my experiments, a connection limit larger than `MaxIdleConns` resulted in few connections piling up, but in a rate the system could handle well (with other words: they went through their `TIME_WAIT` window and were removed from the connection list without the system getting close to the ephemeral connection limit).

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -33,6 +33,7 @@ func ConfigFromURL(awsURL *url.URL) (*aws.Config, error) {
 				}).DialContext,
 				MaxIdleConns:          100,
 				IdleConnTimeout:       90 * time.Second,
+				MaxConnsPerHost:       100,
 				MaxIdleConnsPerHost:   100,
 				TLSHandshakeTimeout:   3 * time.Second,
 				ExpectContinueTimeout: 1 * time.Second,


bboreham:
This code is used for more than S3. Given that it is a highly-reused library I would prefer to see parallelism limited at source, rather than deep in the SDK code.

Worst case the limit should be configurable, off by default, and commented.

JensErat:
As already mentioned, I'm not really sure this is an all-valid default. Making it configurable would make a bunch of use cases much easier, though. Sadly, modifying the HTTP client's from outside the function seems not really possible (or I was just lost in a hell of type assertions). Optionally passing an HTTP client or transport might be somewhat feasible though.

As already discussed in the Loki issue, Loki/Cortex already implemented an improved schema some months ago (which was the new recommended one just few days after we digged into the issue). So on the client-side, there definitely is some improvement works.

When reading through this comment again I have the feeling one of those two occurrences of `MaxIdleConnsPerHost` should actually be `MaxIdleConns`?

```go
// Use a custom http.Client with the golang defaults but also specifying
// MaxIdleConnsPerHost because of a bug in golang https://github.com/golang/go/issues/13801
// where MaxIdleConnsPerHost does not work as expected.
```

bboreham:
> one of those two occurrences of MaxIdleConnsPerHost should actually be MaxIdleConns?

Assuming you mean "one of those two occurrences in the comment", I don't think so.  The struct is  copy-pasted from the Go `DefaultTransport`, plus a line specifying `MaxIdleConnsPerHost`, because (reason).

bboreham:
@JensErat are you planning to come back to this?

PR #187, opened 2020-04-20
Log errors when write fails, not 200; requires Go 1.13

gouthamve:
See discussion here: https://github.com/cortexproject/cortex/pull/2483#discussion_r411273751

Requires projects to use Go 1.13 and above

Signed-off-by: Goutham Veeramachaneni <gouthamve@gmail.com>

### DIFF
--- a/middleware/logging.go
+++ b/middleware/logging.go
@@ -39,7 +39,12 @@ func (l Log) Wrap(next http.Handler) http.Handler {
 		var buf bytes.Buffer
 		wrapped := newBadResponseLoggingWriter(w, &buf)
 		next.ServeHTTP(wrapped, r)
-		statusCode := wrapped.statusCode
+
+		statusCode, err := wrapped.statusCode, wrapped.writeError
+		if err != nil {
+			l.logWithRequest(r).Warnf("%s %s %s, error: %s ws: %v; %s", r.Method, uri, time.Since(begin), err, IsWSHandshakeRequest(r), headers)
+			return
+		}
 		if 100 <= statusCode && statusCode < 500 || statusCode == http.StatusBadGateway || statusCode == http.StatusServiceUnavailable {
 			l.logWithRequest(r).Debugf("%s %s (%d) %s", r.Method, uri, statusCode, time.Since(begin))
 			if l.LogRequestHeaders && headers != nil {
--- a/middleware/logging.go
+++ b/middleware/logging.go
@@ -2,6 +2,8 @@ package middleware
 
 import (
 	"bytes"
+	"context"
+	"errors"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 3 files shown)


middleware/logging.go
```diff
@@ -39,7 +39,12 @@ func (l Log) Wrap(next http.Handler) http.Handler {
 		var buf bytes.Buffer
 		wrapped := newBadResponseLoggingWriter(w, &buf)
 		next.ServeHTTP(wrapped, r)
-		statusCode := wrapped.statusCode
+
+		statusCode, err := wrapped.statusCode, wrapped.writeError
+		if err != nil {
+			l.logWithRequest(r).Warnf("%s %s %s, error: %s ws: %v; %s", r.Method, uri, time.Since(begin), err, IsWSHandshakeRequest(r), headers)
```
bboreham, line 45: probably don't want a warning every time the user's client times out

middleware/logging.go
```diff
@@ -39,7 +39,12 @@ func (l Log) Wrap(next http.Handler) http.Handler {
 		var buf bytes.Buffer
 		wrapped := newBadResponseLoggingWriter(w, &buf)
 		next.ServeHTTP(wrapped, r)
-		statusCode := wrapped.statusCode
+
+		statusCode, err := wrapped.statusCode, wrapped.writeError
```
bboreham, line 43: `err` isn't adding much here

bboreham:
Also: test?

middleware/logging.go
pracucci: Agree. We could log with debug level if `grpc.IsCanceled()` and warning otherwise. Also, I think `grpc.IsCanceled()` is not fully correct because the error could be wrapped. We should improve it like this:

```
func IsCanceled(err error) bool {
	if errors.Is(err, context.Canceled) {
		return true
	}

```

gouthamve:
Added a test and also incorporated the feedback.

> how do they appear in metrics and tracing, after this change?

It would be the same behavior as before unfortunately. `resp.StatusCode == 200` even if a write errors. We only modified the logging middleware and tbh, I am not sure how to catch this in tracing/metrics.

middleware/response.go
```diff
@@ -20,6 +20,7 @@ type badResponseLoggingWriter struct {
 	logBody       bool
 	bodyBytesLeft int
 	statusCode    int
+	writeError    error // The error returned when downstream Write() fails.
```
bboreham, line 23: Strictly this is the _last_ error.  I haven't investigated whether this matters.

middleware/logging.go
```diff
@@ -39,7 +41,18 @@ func (l Log) Wrap(next http.Handler) http.Handler {
 		var buf bytes.Buffer
 		wrapped := newBadResponseLoggingWriter(w, &buf)
 		next.ServeHTTP(wrapped, r)
-		statusCode := wrapped.statusCode
+
+		statusCode, writeErr := wrapped.statusCode, wrapped.writeError
+
+		if writeErr != nil {
+			if errors.Is(writeErr, context.Canceled) {
```
bboreham, line 48: I think this forces consumers to upgrade to Go 1.13, which should be noted

gouthamve: Hrmm, yeah. But where should I note it though?

bboreham: In the PR description.
Be nice if it was self-contained rather than making people read something else.

middleware/response.go
gouthamve: Yeah, you don't typically call `Write()` again after a previous call fails, so it should be fine?

gouthamve:
Tested this by vendoring it into Cortex and reverting  cortexproject/cortex#2483

Before:
```
2020-05-12 16:14:31.381381 I | http: superfluous response.WriteHeader call from github.com/opentracing-contrib/go-stdlib/nethttp.(*statusCodeTracker).WriteHeader (status-code-tracker.go:19)
level=warn ts=2020-05-12T15:14:31.381571716Z caller=logging.go:49 traceID=1bce8fc8c1a57be7 msg="GET /api/v1/query?query=up[10000m]&start=1557759455&step=126144 (500) 729.261961ms Response: \"write tcp 127.0.0.1:9091->127.0.0.1:54276: write: broken pipe\\n\" ws: false; Accept: */*; User-Agent: curl/7.70.0; "
```


After:
```
level=warn ts=2020-05-12T15:13:17.161683485Z caller=logging.go:51 traceID=604b1b0b5218a0e7 msg="GET /api/v1/query?query=up[10000m]&start=1557759455&step=126144 728.757586ms, error: write tcp 127.0.0.1:9091->127.0.0.1:54270: write: broken pipe ws: false; Accept: */*; User-Agent: curl/7.70.0; "
```

The new log is more accurate (the status code in this case is not 500), and also doesn't log the duplicate header write golang error.

PR #189, opened 2020-04-23
Expose server HTTP middleware for custom integrations

pracucci:
We have an issue in Cortex (Prometheus API exposed by the querier are wrapped anymore with the weaveworks/common HTTP middleware). Exposing the HTTP middleware of the server would allow us to easily fix it (see fix here: https://github.com/cortexproject/cortex/pull/2511).

/cc @jtlisi @pstibrany

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -116,10 +116,11 @@ type Server struct {
 	grpcListener net.Listener
 	httpListener net.Listener
 
-	HTTP       *mux.Router
-	HTTPServer *http.Server
-	GRPC       *grpc.Server
-	Log        logging.Interface
+	HTTP           *mux.Router
+	HTTPServer     *http.Server
+	HTTPMiddleware middleware.Interface
+	GRPC           *grpc.Server
+	Log            logging.Interface
 }
 
 // New makes a new Server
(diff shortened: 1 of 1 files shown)


bboreham:
I don't think this is right.
You're trying to expose a structure created as follows:
```
	httpMiddlewares := []middleware.Interface{
		middleware.Tracer{			middleware.Tracer{
			RouteMatcher: router,				RouteMatcher: router,
		},			},
		middleware.Log{			middleware.Log{
			Log: log,				Log: log,
		},			},
		middleware.Instrument{			middleware.Instrument{
			Duration:     requestDuration,				Duration:     requestDuration,
			RouteMatcher: router,				RouteMatcher: router,
		},			},
	}		}
```

The `router` is used as a `RouteMatcher`: that means, logically, this middleware can only be used against the same router.  If used against a different router, which does not have the same paths, the traces and metrics will get reported against `"other"`.

Since `log` can be passed in, the only obstacle to you creating your own similar middleware stack is the `requestDuration` `HistogramVec`.  How about changing `common/server/New()` to accept a `HistogramVec` in `server.Config`, then you can create it externally and use it twice?

pracucci:
Thanks @bboreham for taking the time to go into details. I didn't realise it at the beginning, but after you "your fix is a bug" I tried to understand what you did mean and I finally did. I agree with your conclusion and that's why I closed this PR.

> How about changing common/server/New() to accept a HistogramVec in server.Config, then you can create it externally and use it twice?

Agree.

PR #191, opened 2020-04-30
Added option to disable signal handling by Server.

pstibrany:
In Cortex, we currently have extra logic [(as explained here)](https://github.com/cortexproject/cortex/pull/2542#discussion_r417884587) to handle the fact that `server.Server` type handles SIGINT/SIGTERM signals, that should trigger shutdown of entire server.

This PR adds ability to disable signal handling by Server. This will make it possible to handle signal by Cortex `Run` method instead, and to trigger shutdown of all services (incl. `Server`) cleanly, without doing [special hacks](https://github.com/cortexproject/cortex/blob/master/pkg/cortex/modules.go#L529-L534) for "Server" service. 

Option to disable signal handling is deliberately "negative" ("disable", not "enable"), to keep existing code working as before.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -31,7 +31,8 @@ import (
 	"github.com/weaveworks/common/signals"
 )
 
-type SignalsHandler interface {
+// SignalHandler used by Server.
+type SignalHandler interface {
 	// Starts the signals handler. This method is blocking, and returns only after signal is received,
 	// or "Stop" is called.
 	Loop()
@@ -79,7 +80,7 @@ type Config struct {
 	Log      logging.Interface `yaml:"-"`
 
 	// If not set, default signal handler is used.
-	SignalHandler SignalsHandler `yaml:"-"`
+	SignalHandler SignalHandler `yaml:"-"`
 
 	PathPrefix string `yaml:"http_path_prefix"`
 }
@@ -124,7 +125,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 // Servers will be automatically instrumented for Prometheus metrics.
 type Server struct {
 	cfg          Config
-	handler      SignalsHandler
+	handler      SignalHandler
 	grpcListener net.Listener
 	httpListener net.Listener
 
--- a/server/server.go
+++ b/server/server.go
@@ -66,6 +66,8 @@ type Config struct {
 	GRPCServerTime                  time.Duration `yaml:"grpc_server_keepalive_time"`
 	GRPCServerTimeout               time.Duration `yaml:"grpc_server_keepalive_timeout"`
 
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -258,11 +261,17 @@ func New(cfg Config) (*Server, error) {
...
+		handler = signals.NewHandler(log)
+	}
+
 	return &Server{
 		cfg:          cfg,
 		httpListener: httpListener,
 		grpcListener: grpcListener,
-		handler:      signals.NewHandler(log),
+		handler:      handler,
+		quit:         make(chan struct{}),
```
annanay25, line 274: Would it be better to make the changes to `signals.Handler` with a new method like `handler.Watch()`? (or possibly a better name 😄)

pstibrany: I'm not sure sure that signal handler that doesn't handle signals is a way to go.

pstibrany:
Any chance of getting this merged?

We would like to know whether we can simplify the design of Cortex modules proposal (if this is accepted), or not.

Thank you very much!

server/server.go
```diff
@@ -277,18 +286,28 @@ func RegisterInstrumentation(router *mux.Router) {
...
+		go func() {
+			s.handler.Loop()
+			select {
+			case errChan <- nil:
+			default:
+			}
+		}()
+	} else {
+		go func() {
+			<-s.quit
```
bboreham, line 304: Seems like we have many ways to quit now and I don't know why.
Why do we need Stop vs Shutdown?

pstibrany: `Stop` vs `Shutdown` was there before. `Run` is a blocking method, that returns first error from HTTP/gRPC or nil, if signal has been received or `Stop` was called.
`Shutdown` stops gRPC and HTTP servers.

Since `Stop` works by stopping signal handler, which can no longer be there, I've kept the functionality of `Stop` by using `quit` channel.

If we were doing design from scratch, we could replace blocking `Run` method with non-blocking `Start` (perhaps returning channel with errors), which would make `Stop` unnecessary.

bboreham: Looking further, the signal-handler has _another_ `Stop()` mechanism `SignalReceiver`, which appears to be impossible to use from `Server`.

How about making `handler` something you can pass in config, like `Log`, that has interface `Loop()` and `Stop()`, then Cortex can pass either what you're doing with `quit` or its own actual signal handler?

pstibrany: That sounds good, I'll update PR with your suggestion.

pstibrany:
I've updated PR with interface as per suggestion. I've also modified unrelated `genCerts.sh` to avoid lint complaints (not sure why I'm getting those now).

pstibrany:
Thanks!

PR #193, opened 2020-08-11
Log X-Forwarded-For (or similar) for every request

MichelHollands:
This adds a field with the SourceIPs derived from X-Forwarded-For and the request address to the logging and tracing middleware.  The `server.log-source-ips` setting is used to turn this one. By default it is not turned on.

The `Forwarded`, `X-Real-IP` and `X-Forwarded` headers are examined in this order. If you want to use your own header provide a header name in `server.log-source-ips-header` and a regex to extract it in `server.log-source-ips-regex`.
 
Logs for successful requests are logged at debug level. This is how the log looks like after this change and after the logging level has been set to debug:

```
level=debug ts=2020-08-11T09:46:43.207546Z caller=logging.go:63 traceID=1f208b9870d8cfd1 sourceIPs="1.2.3.4, ::1, ::1" msg="POST /api/v1/push (400) 2.357651ms"
```

A sourceIPs tag is added while tracing:

Screenshot 2020-08-14 at 09 39 15

Signed-off-by: Michel Hollands <michel.hollands@grafana.com>


middleware/forwarded_test.go
```diff
@@ -0,0 +1,100 @@
+package middleware
+
+import (
+	"net/http"
+	"testing"
+)
+
+func TestGetSource(t *testing.T) {
+	type args struct {
```
bboreham, line 9: This struct appears to be adding bulk without adding any value

middleware/forwarded.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// GetSource extracts the X-FORWARDED-FOR header from the given HTTP request
+// and returns a string with it and the remote address
+func GetSource(req *http.Request) string {
+	fwd := req.Header.Get("X-FORWARDED-FOR")
+	if fwd == "" {
+		if req.RemoteAddr == "" {
+			// No X-FORWARDED-FOR header and no RemoteAddr set so we want to return an empty string
+			// Might as well just use the empty string set in req.RemoteAddr
```
bboreham, line 31: `""` is clear and requires no comment.

middleware/forwarded.go
```diff
@@ -0,0 +1,42 @@
...
+	hostStr, _, err := net.SplitHostPort(address)
+	if err != nil {
+		// Invalid IP address, just return it so it shows up in the logs
+		return address
+	}
+	return hostStr
+}
+
+// GetSource extracts the X-FORWARDED-FOR header from the given HTTP request
+// and returns a string with it and the remote address
```
bboreham, line 25: For such a long comment this is inexact.  How about "returns any source addesses we can find in the request, comma-separated"?  Anyone interested in how it is done can read the source.

server/server.go
```diff
@@ -249,17 +256,20 @@ func New(cfg Config) (*Server, error) {
...
-		},
-		middleware.Log{
-			Log: log,
-		},
-		middleware.Instrument{
-			Duration:     requestDuration,
-			RouteMatcher: router,
-		},
+	httpMiddleware := []middleware.Interface{}
+	if !cfg.DisableGeneratedHTTPMiddleware {
```
bboreham, line 260: I think I would put this `if` on the `append` not the static declaration.

server/server_test.go
```diff
@@ -294,6 +295,33 @@ func TestMiddlewareLogging(t *testing.T) {
 	http.DefaultClient.Do(req)
 }
 
+func TestMuxMiddleware(t *testing.T) {
+	var level logging.Level
+	level.Set("info")
+	cfg := Config{
+		HTTPListenAddress: "localhost",
+		HTTPListenPort:    9193,
```
bboreham, line 303: why not put this one on a unique port number rather than moving the ones later on?

or generate unique numbers using `iota`

server/server_test.go
```diff
@@ -294,6 +295,33 @@ func TestMiddlewareLogging(t *testing.T) {
...
+	require.NoError(t, err)
+
+	server.HTTP.HandleFunc("/error500", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(500)
+	})
+
+	go server.Run()
+	defer server.Shutdown()
+
+	req, err := http.NewRequest("GET", "http://127.0.0.1:9192/error500", nil)
```
bboreham, line 320: Why is this port number different to the one in the config?

middleware/logging.go
```diff
@@ -24,6 +24,12 @@ func (l Log) logWithRequest(r *http.Request) logging.Interface {
 		l.Log = l.Log.WithField("traceID", traceID)
 	}
 
+	// Add the source IPs derived from X-Forwarded-For and the request remote address
+	sourceIPs := GetSource(r)
+	if sourceIPs != "" {
+		l.Log = l.Log.WithField("sourceIPs", sourceIPs)
```
bboreham, line 30: Assigning to a member of a pass-by-value parameter strikes me as a bit weird; I see you didn't introduce it but I would prefer you changed this function to use a regular local variable for this purpose.

middleware/forwarded.go
```diff
@@ -0,0 +1,42 @@
...
+// extractHost returns the Host IP address without any port information
+func extractHost(address string) string {
+	hostIP := net.ParseIP(address)
+	if hostIP != nil {
+		return hostIP.String()
+	}
+	var err error
+	hostStr, _, err := net.SplitHostPort(address)
+	if err != nil {
+		// Invalid IP address, just return it so it shows up in the logs
```
bboreham, line 18: What would be an example of an invalid _IP address_ that gets us here?

middleware/forwarded.go
```diff
@@ -0,0 +1,42 @@
...
+			return req.RemoteAddr
+		}
+		return extractHost(req.RemoteAddr)
+	}
+	// If RemoteAddr is empty just return the header
+	if req.RemoteAddr == "" {
+		return fwd
+	}
+	// If both a header and RemoteAddr are present return them both, stripping off any port info from the RemoteAddr
+	return fmt.Sprintf("%v, %v", fwd, extractHost(req.RemoteAddr))
```
bboreham, line 41: You might want to check if the addresses from `RemoteAddr` and `X-Forwarded-For` are identical, in which case only return one.

middleware/forwarded_test.go
MichelHollands: Indeed it is removed.

middleware/forwarded.go
MichelHollands: Changed and the comment is also removed.

middleware/forwarded.go
MichelHollands: Changed as suggested.

server/server_test.go
MichelHollands: Using a unique port number now.

server/server_test.go
MichelHollands: My mistake. It should be the same.

middleware/forwarded.go
MichelHollands: This was added first to catch the proxy making a mistake and returned something invalid like a hostname. The goal of this PR is to add something to diagnose incorrect sending of samples in Cortex. Additionally the `Forwarded` header supports Obfuscated Identifiers (see https://tools.ietf.org/html/rfc7239#section-6.3) which are strings.

middleware/forwarded.go
MichelHollands: Good idea. A test case was also added.

middleware/logging.go
MichelHollands: Using a local variable now.

MichelHollands:
> Thanks for the PR; I made some comments below, mostly stylistic.
> 
> I think I can see three different changes here: removing lines from `go.mod`, adding source logging, and making standard middleware optional. Those should at least be three commits, but I wouldn't expect `go.mod` deletions in a PR like this. Maybe `go.mod` is wrong, but that would be a separate issue.

I've reverted the go.mod and go.sum changes. My mistake about the separate commits. I keep forgetting to do that.

server/server.go
MichelHollands: Makin standard middleware optional was moved out of this PR. So this code is no longer there.

middleware/source_ips.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	} else if fwd := r.Header.Get(xRealIP); fwd != "" {
+		// X-Real-IP should only contain one IP address (the client making the
+		// request).
+		addr = fwd
+	} else if fwd := r.Header.Get(xForwardedFor); fwd != "" {
+		// Only grab the first (client) address. Note that '192.168.0.1,
+		// 10.1.1.1' is a valid key for X-Forwarded-For where addresses after
+		// the first may represent forwarding proxies earlier in the chain.
+		s := strings.Index(fwd, ", ")
```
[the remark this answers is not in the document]
MichelHollands: Good catch. A unit test has been added as well.

middleware/source_ips.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+// Parts copied and changed from gorilla mux proxy_headers.go
+
+var (
+	// De-facto standard header keys.
+	xForwardedFor    = http.CanonicalHeaderKey("X-Forwarded-For")
+	xForwardedHost   = http.CanonicalHeaderKey("X-Forwarded-Host")
+	xForwardedProto  = http.CanonicalHeaderKey("X-Forwarded-Proto")
+	xForwardedScheme = http.CanonicalHeaderKey("X-Forwarded-Scheme")
```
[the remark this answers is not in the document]
MichelHollands: Yes, I got those from gorilla mux but they are not needed here. One of the regexes is also not needed and is also removed.

middleware/source_ips.go
```diff
@@ -0,0 +1,147 @@
...
+	var addr string
+
+	// Use the custom regex only if it was setup
+	if sips.header != "" {
+		hdr := r.Header.Get(sips.header)
+		if hdr == "" {
+			return ""
+		}
+		allMatches := sips.regex.FindAllStringSubmatch(hdr, 1)
+		if allMatches == nil {
```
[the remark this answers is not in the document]
MichelHollands: Yes indeed.

server/server.go
```diff
@@ -120,6 +123,9 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.PathPrefix, "server.path-prefix", "", "Base path to serve all API routes from (e.g. /v1/)")
 	cfg.LogFormat.RegisterFlags(f)
 	cfg.LogLevel.RegisterFlags(f)
+	f.BoolVar(&cfg.LogSourceIPs, "server.log-source-ips", false, "Log the source IPs")
+	f.StringVar(&cfg.LogSourceIPsHeader, "server.log-source-ips-header", "", "Header field storing the source IPs")
+	f.StringVar(&cfg.LogSourceIPsRegex, "server.log-source-ips-regex", "", "Regex for matching the source IPs")
```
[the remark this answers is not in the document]
MichelHollands: Good idea. Added.

server/server.go
```diff
@@ -76,9 +76,12 @@ type Config struct {
 	GRPCServerTime                  time.Duration `yaml:"grpc_server_keepalive_time"`
 	GRPCServerTimeout               time.Duration `yaml:"grpc_server_keepalive_timeout"`
 
-	LogFormat logging.Format    `yaml:"log_format"`
-	LogLevel  logging.Level     `yaml:"log_level"`
-	Log       logging.Interface `yaml:"-"`
+	LogFormat          logging.Format    `yaml:"log_format"`
+	LogLevel           logging.Level     `yaml:"log_level"`
+	Log                logging.Interface `yaml:"-"`
+	LogSourceIPs       bool              `yaml:"log_source_ips"`
```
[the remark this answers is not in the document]
MichelHollands: Yes, that makes a lot of sense. It's changed.

server/server_test.go
```diff
@@ -375,6 +375,33 @@ func TestTLSServer(t *testing.T) {
 	require.EqualValues(t, &empty, grpcRes)
 }
 
+func TestLogSourceIPs(t *testing.T) {
```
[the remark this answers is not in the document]
MichelHollands: Yes it only asserts that the GET request works. I based this on the TestMiddleWareLogging test case above which does the same. 

I've now added a fake logger that checks that the field "sourceIPs" is set on the logger.

server/server.go
```diff
@@ -123,9 +123,9 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&cfg.PathPrefix, "server.path-prefix", "", "Base path to serve all API routes from (e.g. /v1/)")
 	cfg.LogFormat.RegisterFlags(f)
 	cfg.LogLevel.RegisterFlags(f)
-	f.BoolVar(&cfg.LogSourceIPs, "server.log-source-ips", false, "Log the source IPs")
-	f.StringVar(&cfg.LogSourceIPsHeader, "server.log-source-ips-header", "", "Header field storing the source IPs")
-	f.StringVar(&cfg.LogSourceIPsRegex, "server.log-source-ips-regex", "", "Regex for matching the source IPs")
+	f.BoolVar(&cfg.LogSourceIPs, "server.log-source-ips-enabled", false, "Optionally log the source IPs. Default: false")
```
pracucci, line 126: Please don't mention default values in the CLI flag description. When the `--help` is built (or Cortex doc generated), the default values are automatically injected via code.

server/server_test.go
```diff
@@ -397,9 +427,13 @@ func TestLogSourceIPs(t *testing.T) {
 	go server.Run()
 	defer server.Shutdown()
 
+	require.Empty(t, fake.sourceIPs)
+
 	req, err := http.NewRequest("GET", "http://127.0.0.1:9195/error500", nil)
 	require.NoError(t, err)
 	http.DefaultClient.Do(req)
+
+	require.NotEmpty(t, fake.sourceIPs)
```
pracucci, line 436: Instead of testing for `NotEmpty`, the confidence level would be higher if would test for equality on the expected value.

server/server.go
MichelHollands: Good catch. The default is no longer part of the message string.

server/server_test.go
MichelHollands: The test was changed to check for the IP address.

bboreham:
Can you explain a bit more where you envisage the optional header and regex flags being useful?

MichelHollands:
> Can you explain a bit more where you envisage the optional header and regex flags being useful?

Some enterprise customers use non-standard headers (ie different from Forwarded, X-Real-IP and X-Forwarded-For) in their custom reverse proxies. Using a regex as a way to configure this was suggested by Dee Kitchen. He used to work at Cloudflare where they used this successfully.

An added benefit is that customers can configure the headers themselves. Some customers are not keen on giving details about their internal proxies or any internal networking setup for that matter.

bboreham (approved):
I fear this feature is too complicated for a library like this, but I cannot spot any more flaws and it's behind a flag so I'll merge it.
Thanks!

PR #194, opened 2020-08-14
Make HTTP middleware optional

MichelHollands:
Add an option to change the Router and an option to not add the default middleware.

Signed-off-by: Michel Hollands <michel.hollands@grafana.com>

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -62,12 +62,12 @@ type Config struct {
 	HTTPServerWriteTimeout        time.Duration `yaml:"http_server_write_timeout"`
 	HTTPServerIdleTimeout         time.Duration `yaml:"http_server_idle_timeout"`
 
-	GRPCOptions                    []grpc.ServerOption            `yaml:"-"`
-	GRPCMiddleware                 []grpc.UnaryServerInterceptor  `yaml:"-"`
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -276,7 +285,10 @@ func New(cfg Config) (*Server, error) {
 			RouteMatcher: router,
 		},
 	}
-
+	httpMiddleware := []middleware.Interface{}
+	if !cfg.DoNotAddDefaultHTTPMiddleware {
+		httpMiddleware = append(httpMiddleware, defaultHTTPMiddleware...)
```
bboreham, line 290: this `append` seems overkill; a simple assignment would also work?

server/server.go
```diff
@@ -276,7 +285,10 @@ func New(cfg Config) (*Server, error) {
...
+	}
 	httpMiddleware = append(httpMiddleware, cfg.HTTPMiddleware...)
```
bboreham, line 292: or you could put the `DoNotAddDefaultHTTPMiddleware` condition on this line, so you either do an append or take `cfg.HTTPMiddleware` unaltered

MichelHollands: That indeed reads better.

bboreham (approved):
Thanks.

Issue #192, opened 2020-07-13
HTTP Middleware Error Logging: Vision on StatusBadGateway/ServiceUnavailable


joe-elliott:
The http logging middleware splits out different request results and logs them as either debug or warn.  Generally errors are logged as warn and successes are logged as debug.

https://github.com/weaveworks/common/blob/4b1847531bc94f54ce5cf210a771b2a86cd34118/middleware/logging.go#L56-L64

We need to log the below error conditions that are currently being logged as debug.  Unfortunately, due to volume, we can't turn on debug logging.
```
statusCode == http.StatusBadGateway || statusCode == http.StatusServiceUnavailable
```

My guess is that these above status's are bundled into debug logging due to volume if the backend is unavailable.  We would like to log these failures above debug logging, but also recognize that the volume would be too great to log if a backend is down. 

The change we'd like to make:
- Move `http.StatusBadGateway` and `http.StatusServiceUnavailable` to be logged with the other errors
- Use a configurable rate limited logger to log errors instead of logging 100% of all errors at Warn

Thoughts?

bboreham:
Background is here: https://github.com/cortexproject/cortex/issues/810, http://github.com/weaveworks/common/pull/84

I would be ok with sampling the messages (we have `-event.sample-rate` in cortex already); I guess a rate limit is also fine

You could also sample after the line hits the logfile?

joe-elliott:
Thanks for the background.  As suspected those two error codes were just overwhelming the logs and so they got removed.  It sounds like you're ok with the general idea so I will submit a PR and we can discuss details there.

> You could also sample after the line hits the logfile?
Unsure what you mean by this.  Like log everything but only push a certain subset of the logfile to the backend?

PR #195, opened 2020-08-14
Rate limited errors

joe-elliott:
This PR attempts to address concerns in #192.  It provides

- A rate limited logger
- A method of configuring the logger for high volume http errors through the server
- Fixed possible race condition

While working on this PR I noticed this:
```
func (l Log) logWithRequest(r *http.Request) logging.Interface {
	traceID, ok := ExtractTraceID(r.Context())
	if ok {
		l.Log = l.Log.WithField("traceID", traceID)
	}

	return user.LogWith(r.Context(), l.Log)
}
```
Particularly the line:
`l.Log = l.Log.WithField("traceID", traceID)`
I'm unsure why the struct var is being overwritten in `logWithRequest`.  This seems like a possible race condition to me?  I assume there's one instance of the middleware that is shared by all requests in the pipeline?


bboreham:
> why the struct var is being overwritten

it's a local variable, so not likely to cause a problem, but this was made less confusing in #193

middleware/logging.go
```diff
@@ -53,18 +58,31 @@ func (l Log) Wrap(next http.Handler) http.Handler {
...
-		if 100 <= statusCode && statusCode < 500 || statusCode == http.StatusBadGateway || statusCode == http.StatusServiceUnavailable {
+		if 100 <= statusCode && statusCode < 500 {
 			l.logWithRequest(r).Debugf("%s %s (%d) %s", r.Method, uri, statusCode, time.Since(begin))
 			if l.LogRequestHeaders && headers != nil {
 				l.logWithRequest(r).Debugf("ws: %v; %s", IsWSHandshakeRequest(r), string(headers))
 			}
+		} else if statusCode == http.StatusBadGateway || statusCode == http.StatusServiceUnavailable {
+			l.logHighVolumeError(r, "%s %s (%d) %s", r.Method, uri, statusCode, time.Since(begin))
+			if l.LogRequestHeaders && headers != nil {
+				l.logHighVolumeError(r, "ws: %v; %s", IsWSHandshakeRequest(r), string(headers))
```
bboreham, line 69: do you think we want this on a different line to the method?
(I see it did this in the code before, but only at debug level)

joe-elliott: i think it's ok.  this way the user can configure `LogRequestHeaders` and `HighVolumeErrorLog` independently and any combination works.

bboreham: What I'm thinking is: suppose you want to look at logs for a specific request; you can grep for the URI.  But the headers are on the next line, which is awkward.

joe-elliott: oh, that's a good point.  i hadn't considered that.  i'll push a small change to unify them

logging/ratelimit.go
```diff
@@ -0,0 +1,79 @@
...
+	return &rateLimitedLogger{
+		next:    l.next.WithField(key, value),
+		limiter: rate.NewLimiter(l.limiter.Limit(), 0),
+	}
+}
+
+func (l *rateLimitedLogger) WithFields(f Fields) Interface {
+	return &rateLimitedLogger{
+		next:    l.next.WithFields(f),
+		limiter: rate.NewLimiter(l.limiter.Limit(), 0),
```
bboreham, line 77: Why do we get a new limiter here?
Also, does it work with a burst size of zero?

joe-elliott: Yes, the burst limit was wrong.  Added a test and fixed.  As far as the new limiter I figured calling `WithFields()` created a new logger that should have an independent rate limit, but I can go either way on this.

bboreham: Thing is, code often calls `WithFields()` on every operation, e.g. the example you gave in the description, or this one:
https://github.com/weaveworks/common/blob/61ffdd448099f94ec3dc9aaf93705399b97a2bb2/middleware/grpc_logging.go#L30

I expect the case you're trying to hit is where a high volume of logs from from lots of operations, so you would want the same rate-limit to be applied across them all.  Try it!

joe-elliott: You're right.  Thanks man :+1: 

Fixed.

bboreham:
Sorry I haven't been back here for a while.  Code looks ok now; do you still want to merge it?
Have you tried it out for real?

bboreham:
@joe-elliott ping on questions

joe-elliott:
Honestly, this has kind of fallen off of my radar.  The initial need for this no longer something I am no longer involved in.  Having said that I do think the rate limited logger has some value perhaps in weaveworks/common if we want to keep it. 

The original reason this existed was to catch errors returned from Loki in the cortex-gw when the query frontend crashed.  @cyriltovena @owen-d what do you think?  is there still a need for this?

bboreham:
Right now I have a high rate of errors like this:

```
level=warn ts=2021-03-01T11:35:02.01609324Z caller=grpc_logging.go:38 method=/cortex.Ingester/Push duration=2.741131ms err="rpc error: code = Code(429) desc = user=redacted: per-metric series limit (local limit: 300000 global limit: 0 actual local limit: 300000) exceeded for series {redacted}" msg="gRPC\n"
```

I think this PR as it stands would not help me, because it is focused on one specific http error.

PR #196, opened 2020-08-17
Track message size for HTTP and gRPC

pstibrany:
This PR adds tracking of message sizes (body size in case of HTTP request, individual messages for gRPC calls) and exports them as histograms per method and route (or method name for gRPC).

Note that this is similar to approach to https://github.com/cortexproject/cortex/pull/3036, except that PR only addresses gRPC part. While addressing HTTP part, it became obvious that it should perhaps go to common library instead.

This PR uses https://github.com/felixge/httpsnoop/ for wrapping HTTP response when calling the next handler. Httpsnoop handles more interfaces, where previously only `http.Hijacker` was supported by wrapper.


pstibrany:
Note that this is similar to approach to https://github.com/cortexproject/cortex/pull/3036, except that PR only addresses gRPC part. While addressing HTTP part, it became obvious that it should perhaps go to common library instead.

pstibrany:
Update: also added gauge for tracking number of inflight requests per route.

tomwilkie:
LGTM thanks Peter!

middleware/instrument.go
```diff
@@ -42,16 +48,28 @@ func IsWSHandshakeRequest(req *http.Request) bool {
...
-		)
-		i.Duration.WithLabelValues(r.Method, route, status, isWS).Observe(took.Seconds())
+		inflight := i.InflightRequests.WithLabelValues(r.Method, route)
+		inflight.Inc()
+		defer inflight.Dec()
+
+		origBody := r.Body
+		defer func() {
+			r.Body = origBody
+		}()
```
pracucci, line 59: I would add a comment to explain why you restore it.

middleware/grpc_stats.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	switch s := rpcStats.(type) {
+	case *stats.Begin:
+		// Ignored.
+	case *stats.End:
+		// Ignored.
+	case *stats.InHeader:
+		// WireLength seems to be incorrect for large headers -- it uses
+		// length of last frame (16K) even for headers in megabytes.
```
bboreham, line 50: Please link the gRPC issue number.

middleware/grpc_stats_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"google.golang.org/grpc/health/grpc_health_v1"
+
+	"github.com/weaveworks/common/middleware/middleware_test"
+)
+
+func TestGrpcStats(t *testing.T) {
+	reg := prometheus.NewRegistry()
+
+	received := prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "received_payload_size",
```
bboreham, line 26: should "size" be "bytes" for consistency?

bboreham (commented):
I think the commit message "Added metrics for http body sizes." should mention adding a new library.  
Or be split into two commits, one of which changes the implementation and one adds the metrics.

middleware/grpc_stats.go
pstibrany: I've removed the comment instead. I couldn't find gRPC issue. It's something I've observed, but I'm not even sure if it's an issue or not (I would expect WireLength to be higher than message size, but it was just 16K).

server/server_test.go
```diff
@@ -228,6 +231,156 @@ func TestErrorInstrumentationMiddleware(t *testing.T) {
...
+		request_message_bytes_bucket{method="GET",route="error500",le="2.62144e+06"} 1
+		request_message_bytes_bucket{method="GET",route="error500",le="5.24288e+06"} 1
+		request_message_bytes_bucket{method="GET",route="error500",le="1.048576e+07"} 1
+		request_message_bytes_bucket{method="GET",route="error500",le="2.62144e+07"} 1
+		request_message_bytes_bucket{method="GET",route="error500",le="5.24288e+07"} 1
+		request_message_bytes_bucket{method="GET",route="error500",le="1.048576e+08"} 1
+		request_message_bytes_bucket{method="GET",route="error500",le="2.62144e+08"} 1
+		request_message_bytes_bucket{method="GET",route="error500",le="+Inf"} 1
+		request_message_bytes_sum{method="GET",route="error500"} 0
+		request_message_bytes_count{method="GET",route="error500"} 1
```
bboreham, line 318: Not your fault, but this is ugly as heck.  Suggest `prometheus/testutil` could be augmented with a way to provide expected histogram counts.

pstibrany: We can make it slightly more beautiful by not using scientific notation in bucket sizes. Not sure if that would help much though.

bboreham: Is that something you can control?  

I was thinking something like `testutil.ExpectedHistogram(inflightRequests, "GET", "error500", 100)`

pstibrany: We can control format of the floats in our text output, yet.

As somewhat nicer but still ugly solution, I'm thinking
```go
		# HELP request_message_bytes 
		# TYPE request_message_bytes histogram`+
		histogram([]float64{0}, middleware.BodySizeBuckets, "request_message_bytes_bucket", "method", "GET", "route", "error500")+
		histogram([]float64{4}, middleware.BodySizeBuckets, "request_message_bytes_bucket", "method", "GET", "route", "sleep10")+
		histogram([]float64{0}, middleware.BodySizeBuckets, "request_message_bytes_bucket", "method", "GET", "route", "succeed")+...
```

with `histogram` function to generate the text.

Your proposed `testutil.ExpectedHistogram` is an option too. We can try it in this lib first, and if it is nice, can try to send to Prometheus client library.

bboreham (approved):
lgtm, thanks!

server/server.go
```diff
@@ -191,6 +191,29 @@ func New(cfg Config) (*Server, error) {
 	}, []string{"method", "route", "status_code", "ws"})
 	prometheus.MustRegister(requestDuration)
 
+	receivedMessageSize := prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: cfg.MetricsNamespace,
+		Name:      "request_message_bytes",
+		Help:      "Size (in bytes) of messages received in the request.",
+		Buckets:   middleware.BodySizeBuckets,
```
bboreham, line 198: Today I tried to use this metric, and found that all requests were in the smallest bucket of 1MB.
Was there some expectation that servers would receive massive requests?

pstibrany: The idea of this work was to find out what the right limit for gRPC message sizes is. That's why we used relatively big buckets.

PR #198, opened 2020-08-28
Fixed aws.ConfigFromURL() when URL contains @ but no user/pass

pracucci:
Starting from an issue opened in Cortex https://github.com/cortexproject/cortex/issues/3034, I've realised an edge case in `aws.ConfigFromURL()`. If the URL is in the format `scheme://@host/path`, the static credentials are configured in the S3 config but with an empty ID / secret, because no user/pass is specified in the URL before the `@`.

In this PR I'm proposing a fix for that.

### DIFF
--- a/aws/config_test.go
+++ b/aws/config_test.go
@@ -25,6 +25,13 @@ func TestAWSConfigFromURL(t *testing.T) {
 			"http://s3.default.svc.cluster.local:4569",
 		},
 		{
+			"s3://@us-east-1/test-bucket",
+			"",
+			"",
+			"us-east-1",
+			"",
+		},
+		{
 			"dynamodb://user:pass@dynamodb.default.svc.cluster.local:8000/cortex",
 			"user",
 			"pass",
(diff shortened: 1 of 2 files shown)


aws/config_test.go
```diff
@@ -54,10 +61,10 @@ func TestAWSConfigFromURL(t *testing.T) {
 			cfg, err := ConfigFromURL(parsedURL)
 			require.NoError(t, err)
 
-			if cfg.Credentials == nil {
-				assert.Equal(t, "", tc.expectedKey)
-				assert.Equal(t, "", tc.expectedSecret)
+			if tc.expectedKey == "" && tc.expectedSecret == "" {
+				assert.Nil(t, cfg.Credentials)
```
pracucci, line 65: I changed this assertion to ensure that `cfg.Credentials` are not set at all when we expect no key and secret.

aws/config_test.go
```diff
@@ -70,6 +77,8 @@ func TestAWSConfigFromURL(t *testing.T) {
 			if tc.expectedEp != "" {
 				require.NotNil(t, cfg.Endpoint)
 				assert.Equal(t, tc.expectedEp, *cfg.Endpoint)
+			} else {
+				assert.Nil(t, cfg.Endpoint)
```
pracucci, line 81: If we expect no endpoint, we should make sure it's not set.

tomwilkie:
LGTM thanks Marco!

bboreham (approved):
LGTM, thanks!

PR #200, opened 2020-09-04
s/Kuberneretes/Kubernetes/

jdbaldry:
Signed-off-by: Jack Baldry <jack.baldry@grafana.com>


dholbach (approved):
Thanks for spotting this!

PR #201, opened 2020-10-23
Support multiple org IDs

simonswine:
This adds support for multiple org IDs to allow implementation of https://github.com/cortexproject/cortex/pull/3250

Signed-off-by: Christian Simon <simon@swine.de>

### DIFF
--- a/user/grpc.go
+++ b/user/grpc.go
@@ -1,6 +1,8 @@
 package user
 
 import (
+	"reflect"
+
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/metadata"
 )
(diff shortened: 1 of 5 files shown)


bboreham:
Not sure I want this in the core path.
What are the trade-offs if we put multiple IDs somewhere else?

simonswine:
Fair enough. 

Assuming you mean with with "somewhere else" downstream projects (e.g. Cortex in this case).

This would require some work to make sure we can put the `[Inject|Extract]-[GRCP|HTTP]` methods behind an interface, to able to customise or we they would need to be moved/reimplemented downstream as well:
- https://github.com/weaveworks/common/blob/313edde7a4559925dddf2aa97185cf519ef2642d/middleware/http_auth.go#L12
- https://github.com/weaveworks/common/blob/313edde7a4559925dddf2aa97185cf519ef2642d/middleware/grpc_auth.go#L11

bboreham:
Even within Cortex I am unclear that we'd want the ingest path to accept multiple IDs.
But I'm still interested in the trade-offs (like, which things get harder, or slower).

simonswine:
I think it would be really great getting your thoughts and suggestions on this proposal: https://github.com/cortexproject/cortex/pull/3364. (I really should have linked it here)

This draft PR, was mainly used to trial a couple of approaches and share it with others to find the best approach.

PR #202, opened 2020-10-27
Add exemplar support to middleware

beorn7:
@joe-elliott as discussed, this is my attempt of a PR. Please verify that this still works. I had to jump through a few hoops to resolve the merge conflicts.

@bboreham for context: As just demoed during Grafana Lab's Observability Con, we are starting to utilize the exemplar support in Prometheus and OpenMetrics. This PR activates the exposition of exemplars with the Jaeger Trace ID if there is one.


middleware/http_tracing.go
```diff
@@ -54,3 +54,19 @@ func ExtractTraceID(ctx context.Context) (string, bool) {
...
+	sp := opentracing.SpanFromContext(ctx)
+	if sp == nil {
+		return "", false
+	}
+	sctx, ok := sp.Context().(jaeger.SpanContext)
+	if !ok {
+		return "", false
+	}
+
+	return sctx.TraceID().String(), sctx.IsSampled() && sctx.IsSamplingFinalized()
```
joe-elliott, line 71: I would drop `IsSamplingFinalized()`.  The code paths that set sampling "finalized" or not are complex.  I think for almost all scenarios just `IsSampled()` is sufficient and less prone to complex edge cases.

joe-elliott (approved):
Thanks @beorn7 looks awesome!

middleware/instrument.go
```diff
@@ -68,9 +68,20 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
 			next.ServeHTTP(ww, r)
 		})
 
-		i.Duration.WithLabelValues(r.Method, route, strconv.Itoa(respMetrics.Code), isWS).Observe(respMetrics.Duration.Seconds())
 		i.RequestBodySize.WithLabelValues(r.Method, route).Observe(float64(rBody.read))
 		i.ResponseBodySize.WithLabelValues(r.Method, route).Observe(float64(respMetrics.Written))
+
+		if traceID, ok := ExtractSampledTraceID(r.Context()); ok {
+			i.Duration.WithLabelValues(r.Method, route, strconv.Itoa(respMetrics.Code), isWS).
```
bboreham, line 75: suggest a different structure to avoid repeating parts:
```
observer := i.Duration.WithLabelValues(r.Method, route, strconv.Itoa(respMetrics.Code), isWS)
if ... {
    observer.(prometheus.ExemplarObserver).ObserveWithExemplar()
} else {
    observer.Observe()
}
```

bboreham (commented):
Thanks!  

I wonder if it would be worth just unilaterally changing ExtractTraceID() to do what you did - why would someone want the trace id if it's not sampled?

beorn7:
> I wonder if it would be worth just unilaterally changing ExtractTraceID() to do what you did - why would someone want the trace id if it's not sampled?

That's a good question. As long as we don't know, I guess we have to assume there are users of `ExtractTraceID()` out there that are interested in non-sampled trace IDs. (I'm not a tracing expert and have no strong opinion here.)

middleware/http_tracing.go
beorn7: If you say so, happy to do so.

middleware/instrument.go
beorn7: Done. Thanks.

beorn7:
Thanks, @bboreham and @joe-elliott . I addressed your comments.

middleware/instrument.go
```diff
@@ -68,9 +68,20 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
...
 		i.RequestBodySize.WithLabelValues(r.Method, route).Observe(float64(rBody.read))
 		i.ResponseBodySize.WithLabelValues(r.Method, route).Observe(float64(respMetrics.Written))
+
+		histogram := i.Duration.WithLabelValues(r.Method, route, strconv.Itoa(respMetrics.Code), isWS)
+		if traceID, ok := ExtractSampledTraceID(r.Context()); ok {
+			// Need to type-convert the Observer to an
+			// ExemplarObserver. This will always work for a
+			// HistogramVec.
+			histogram.(prometheus.ExemplarObserver).ObserveWithExemplar(
+				respMetrics.Duration.Seconds(), prometheus.Labels{"traceID": traceID},
```
bboreham, line 80: Is it OK to hard-code `"traceID"` ?   I didn't find the documentation on `ObserveWithExemplar()` very helpful in understanding what those labels are used for.

bboreham (approved):
lgtm.  One question below, and a thought: I guess we will have to port all this to OpenTelemetry one day.

middleware/instrument.go
beorn7: From the perspective of prometheus/client_golang, these are completely free-form, and it's totally up to the user what to use them for. We are using them for trace IDs, but others might want to point to a log entry. Or they might want to put in an additional parameter for particular events, e.g. the query string for a query that resulted in an exceptionally long processing time. We are still at the beginning of the exemplar journey, so we have to see what they will be used for.

beorn7: So why "traceID" here in particular? That's just the convention that we came up with. As we are the first doing it in the Prometheus space, we have the power of naming things… Despite hardcoding it here, it might still change. People will come up with various bikeshedding arguments (e.g. it's more common in Prometheus land to use snake case, so some might argue it should be `trace_id`), and if you prefer, we could make it configurable here.

bboreham: OK let's move forward and change it later if we need to.

bboreham (approved):
lgtm

beorn7:
Thanks @bboreham .

bboreham:
Does anyone recall why the `instrument` package was not similarly extended?

beorn7:
I guess back then, this change was just an experiment to play with exemplars in this particular case.

PR #204, opened 2020-11-05
tracing: Allow specifying JAEGER_ENDPOINT instead of sampling server or local agent port

### LINKED ISSUE #203: Better handle configuring jaeger tracing with thrift http transport
I'm trying to configure loki to send traces to tempo over the thrift http protocol for for the opentelemetry jaeger receiver. 
Loki uses `weaveworks/common` to set this up.

I've configured my components with the `JAEGER_ENDPOINT` env var, but I'm getting an error:

```
020-11-04T14:53:17-08:00 level=error ts=2020-11-04T22:53:17.150980442Z caller=main.go:112 msg="error in initializing tracing. tracing will not be enabled" err="no trace report agent or config server specified"
```

This appears to come from here:
https://github.com/grafana/loki/blob/master/vendor/github.com/weaveworks/common/tracing/tracing.go#L32-L43

t appears there may be a work-around by specifying some other options, like `JAEGER_AGENT_HOST`, as that will cause the `LocalAgentHostPort` configuration to get a default value, bypassing the validation error. And according to https://github.com/grafana/loki/blob/master/vendor/github.com/uber/jaeger-client-go/config/config.go#L416, if the `CollectorEndpoint` (`JAEGER_ENDPOINT`) is set, then then the UDP transport isn't even used.

I'm still testing the work around, but https://github.com/grafana/tempo/blob/master/example/docker-compose/docker-compose.loki.yaml#L70-L73 indicates it might work.

chancez:
closes #203

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -10,7 +10,7 @@ import (
 
 // ErrInvalidConfiguration is an error to notify client to provide valid trace report agent or config server
 var (
-	ErrBlankTraceConfiguration = errors.New("no trace report agent or config server specified")
+	ErrBlankTraceConfiguration = errors.New("no trace report agent, config server, or collector endpoint specified")
 )
 
 // installJaeger registers Jaeger as the OpenTracing implementation.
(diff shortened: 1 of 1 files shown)


joe-elliott (approved):
Ha, I noticed this behavior, but never tracked down why it occurred.  Good fix.

@bboreham can you review when you get a chance?

bboreham (approved):
Thanks

bboreham:
Ideally this PR should have updated the comment on lines 29-31.

PR #205, opened 2020-11-18
Expose gRPC keepalive enforcement policy options.

pstibrany:
We'd like to modify gRPC keepalive policies on the server, but these options are currently not exposed.

This PR changes that.

Note that default values are consistent with gRPC server options.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -69,14 +69,16 @@ type Config struct {
 	Router                        *mux.Router                    `yaml:"-"`
 	DoNotAddDefaultHTTPMiddleware bool                           `yaml:"-"`
 
-	GPRCServerMaxRecvMsgSize        int           `yaml:"grpc_server_max_recv_msg_size"`
-	GRPCServerMaxSendMsgSize        int           `yaml:"grpc_server_max_send_msg_size"`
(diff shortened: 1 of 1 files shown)


pracucci (approved):
LGTM

bboreham (approved):
Seems plausible.  I don't really understand the use-case for MinTimeBetweenPings but willing to believe there is one.

bboreham:
Actually what is the behaviour now (before this PR) on min time between pings?  If I send at 4min 59sec I get told to go away?

pstibrany:
> Actually what is the behaviour now (before this PR) on min time between pings? If I send at 4min 59sec I get told to go away?

This PR doesn't modify the behaviour (settings are the same as default values used by gRPC).

Server uses 5 min default min time for pings. If client sends another ping within this period, it gets "ping strike". When it reaches 3 ping strikes, client connection is told to go away, and connection is closed.

It's implemented here: https://github.com/cortexproject/cortex/blob/8cfa79c9265492fbd78a3d89d29e526d35e3db9b/vendor/google.golang.org/grpc/internal/transport/http2_server.go#L672

PR #206, opened 2020-11-18
Split setting of default values and registering flags.

pstibrany:
This PR proposes to split setting of default values and flag registration. This allows clients to modify default values, and have correct help message printed, by using:

```
	var cfg Config
	cfg.DefaultValues()
	cfg.RegisterInstrumentation = false
	cfg.RegisterFlags(flag.CommandLine)
```

I'd like to know what you think about this idea.

### DIFF
--- a/logging/format.go
+++ b/logging/format.go
@@ -13,9 +13,12 @@ type Format struct {
 	Logrus logrus.Formatter
 }
 
+func (f *Format) DefaultValues() {
+	f.Set("logfmt")
+}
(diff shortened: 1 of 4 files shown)


bboreham:
I don't think I want this as a breaking change.  I think it could be expressed as a new method with backwards-compatible behaviour on the old one.

pstibrany:
Thanks. Decided not to go forward with this approach.

PR #209, opened 2021-01-06
Uses a better body reader for the http request of the grpc server.

cyriltovena:
It might not be obvious but this small changes has a big impact:

1. if you read the `http.NewRequest` code you will realize that the content lenght is computed only if the body interface is from a specific type.

```go
if body != nil {
		switch v := body.(type) {
		case *bytes.Buffer:
			req.ContentLength = int64(v.Len())
			buf := v.Bytes()
			req.GetBody = func() (io.ReadCloser, error) {
				r := bytes.NewReader(buf)
				return ioutil.NopCloser(r), nil
			}
		case *bytes.Reader:
			req.ContentLength = int64(v.Len())
			snapshot := *v
			req.GetBody = func() (io.ReadCloser, error) {
				r := snapshot
				return ioutil.NopCloser(&r), nil
			}
		case *strings.Reader:
			req.ContentLength = int64(v.Len())
			snapshot := *v
			req.GetBody = func() (io.ReadCloser, error) {
				r := snapshot
				return ioutil.NopCloser(&r), nil
			}
		default:
			// This is where we'd set it to -1 (at least
			// if body != NoBody) to mean unknown, but
			// that broke people during the Go 1.8 testing
			// period. People depend on it being 0 I
			// guess. Maybe retry later. See Issue 18117.
		}

```

This means that until then, the `ContentLength` was not available to downstream handler.

2. I could have simply remove the useless `ioutil.NopCloser` but instead I replace it with `bytes.Buffer`. Now this means we don't need anymore to make a copy of the body if we want to use it. We can just cast it.

For example in Cortex this code https://github.com/cortexproject/cortex/blob/master/pkg/util/http.go#L94-L99 will never allocate properly the buffer and will make multiple allocations call while reading the body because `expectedSize` is always 0.

This happens in Loki too, and most likely in Tempo.

### DIFF
--- a/httpgrpc/server/server.go
+++ b/httpgrpc/server/server.go
@@ -3,6 +3,7 @@ package server
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
@@ -11,7 +12,7 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/grpc-ecosystem/go-grpc-middleware"
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	otgrpc "github.com/opentracing-contrib/go-grpc"
 	"github.com/opentracing/opentracing-go"
 	"github.com/sercand/kuberesolver"
@@ -37,16 +38,35 @@ func NewServer(handler http.Handler) *Server {
 	}
 }
 
+type nopCloser struct {
+	*bytes.Buffer
+}
+
+func (nopCloser) Close() error { return nil }
+
+func nopCloserBuffer(b *bytes.Buffer) io.ReadCloser {
+	return nopCloser{b}
+}
+
+// BodyFromReader attempt to return the body as a bytes buffer to avoid reading and copying one that already exists.
+func BodyFromReader(r io.ReadCloser) (*bytes.Buffer, bool) {
+	if nop, ok := r.(nopCloser); ok {
+		return nop.Buffer, ok
+	}
+	return nil, false
+
+}
+
 // Handle implements HTTPServer.
 func (s Server) Handle(ctx context.Context, r *httpgrpc.HTTPRequest) (*httpgrpc.HTTPResponse, error) {
-	req, err := http.NewRequest(r.Method, r.Url, bytes.NewBuffer(r.Body))
+	req, err := http.NewRequest(r.Method, r.Url, nopCloserBuffer(bytes.NewBuffer(r.Body)))
 	if err != nil {
 		return nil, err
 	}
 	toHeader(r.Headers, req.Header)
 	req = req.WithContext(ctx)
 	req.RequestURI = r.Url
-
+	req.ContentLength = int64(len(r.Body))
 	recorder := httptest.NewRecorder()
 	s.handler.ServeHTTP(recorder, req)
 	resp := &httpgrpc.HTTPResponse{
--- a/httpgrpc/server/server.go
+++ b/httpgrpc/server/server.go
@@ -48,13 +48,12 @@ func nopCloserBuffer(b *bytes.Buffer) io.ReadCloser {
 	return nopCloser{b}
 }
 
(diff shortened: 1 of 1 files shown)


cyriltovena:
I've found an even better way. NewRequest does the wrapping of ioutil.NopCloser if not done, and that makes it impossible to cast back to a buffer.

pracucci (commented):
Interesting! Changes LGTM.

bboreham:
I found it rather hard to follow this code, because it's really doing two different things, and the second commit undoes part of the first commit without explanation. I think I understand what is intended now, after some Slack conversation.

IMHO these should be two PRs, or at least two non-overlapping commits.

The addition of `nopCloser` should be motivated; you declare a similar thing to be "useless" in your description.

Also I think the naming of `BodyFromReader` could better describe what (I think) you are doing - providing fast access to someone who wants a `[]byte` from a `Reader`.  And either explain why the parameter is `ReadCloser` or change it.

httpgrpc/server/server.go
```diff
@@ -37,16 +38,35 @@ func NewServer(handler http.Handler) *Server {
...
+}
+
+func (nopCloser) Close() error { return nil }
+
+func nopCloserBuffer(b *bytes.Buffer) io.ReadCloser {
+	return nopCloser{b}
+}
+
+// BodyFromReader attempt to return the body as a bytes buffer to avoid reading and copying one that already exists.
+func BodyFromReader(r io.ReadCloser) (*bytes.Buffer, bool) {
```
bboreham, line 52: I think the idiomatic Go way to do this would be as a method, which the caller can test via type assertion.

cyriltovena:
> I found it rather hard to follow this code, because it's really doing two different things, and the second commit undoes part of the first commit without explanation. I think I understand what is intended now, after some Slack conversation.

The first commit fixed the `ContentLength` but then I wasn't able to get the `bytes.buffer` like I wanted because see below.

> The addition of `nopCloser` should be motivated; you declare a similar thing to be "useless" in your description.

I had to introduce `nopCloser` because if the body is not a `ReaderCloser` the NewRequest will do the wrapping with io.NopCloser, the problem is that this one is private and it's impossible to get the body without reflection.



See the full NewRequest code. https://golang.org/src/net/http/request.go?s=28366:28467#857

```
	rc, ok := body.(io.ReadCloser)
	if !ok && body != nil {
		rc = ioutil.NopCloser(body)
	}
```

I missed that at first, this is why it's another commit and that's what motivates this new struct.

bboreham:
Please rebase to create a set of commits where each commit is explained by its description and does not undo part of an earlier commit.

httpgrpc/server/server.go
```diff
@@ -37,16 +38,35 @@ func NewServer(handler http.Handler) *Server {
...
+
+func (nopCloser) Close() error { return nil }
+
+func nopCloserBuffer(b *bytes.Buffer) io.ReadCloser {
+	return nopCloser{b}
+}
+
+// BytesBufferFromReader attempts to cast a `io.ReadCloser` as a `bytes.Buffer` to avoid reading and copying one that already exists.
+// In case of failure it will return nil and a false boolean.
+func BytesBufferFromReader(r io.ReadCloser) (*bytes.Buffer, bool) {
```
bboreham, line 53: I think the idiomatic Go way to do this would be as a method, which the caller can test via type assertion.

cyriltovena: In my opinion, I think this is simpler, what you described means I need to know about the type and the method. 

I'll do the change anyway, it'll save time for both of us.

cyriltovena:
Closing in favor of:

- https://github.com/weaveworks/common/pull/211
- https://github.com/weaveworks/common/pull/210

httpgrpc/server/server.go
bboreham: Just the method.  Example: https://github.com/grpc/grpc-go/blob/4be647f/rpc_util.go#L731-L733

PR #210, opened 2021-01-11
Set the content-length of httpgrpc request.

cyriltovena:
http.NewRequest can't set it while using a `io.NopCloser`.


bboreham (approved):
Thanks!

cyriltovena:
@bboreham Do you mind merging this? I've got more use case for this in Loki.

PR #211, opened 2021-01-11
Replace `ioutil.NopCloser` with a custom one.

cyriltovena:
This way we can easily get direct access to the underlaying `bytes.Buffer`. This means we don't need to copy the data anymore.


bboreham (approved):
Thanks!

PR #214, opened 2021-04-13
Add metric "tcp_connections" for number of TCP connections to server.

stevesg:
Counts the number of accepted connections and the number of closed
connections, by intercepting the listener for each port (http/grpc).

Signed-off-by: Steve Simpson <steve.simpson@grafana.com>

### DIFF
--- a/middleware/counting_listener.go
+++ b/middleware/counting_listener.go
@@ -1,0 +1,43 @@
+package middleware
+
+import (
+	"net"
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// CountingListener returns a Listener that increments a Prometheus gauge when
+// a connection is accepted, and decrements the gauge when the connection is closed.
+func CountingListener(l net.Listener, g prometheus.Gauge) net.Listener {
+	return &countingListener{Listener: l, gauge: g}
+}
+
+type countingListener struct {
+	net.Listener
+	gauge prometheus.Gauge
+}
+
+func (c *countingListener) Accept() (net.Conn, error) {
+	conn, err := c.Listener.Accept()
+	if err != nil {
+		return nil, err
+	}
+	c.gauge.Inc()
+	return &countingListenerConn{Conn: conn, gauge: c.gauge}, nil
+}
+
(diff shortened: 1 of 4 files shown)


pracucci (approved):
Thanks @stevesg. For what it's worth, the changes LGTM.

middleware/counting_listener.go
```diff
@@ -0,0 +1,43 @@
...
+	net.Conn
+	gauge prometheus.Gauge
+	once  sync.Once
+}
+
+func (l *countingListenerConn) Close() error {
+	err := l.Conn.Close()
+
+	// Only ever decrement the gauge once in case of badly behaving callers.
+	l.once.Do(func() { l.gauge.Dec() })
```
bboreham, line 40: I can sort of see what you might be getting at, but doesn't this break it for well-behaved callers?

stevesg: The `Once` is per-connection, so we only decrement once per connection closing, which is what we want, unless I'm missing something.

It doesn't look like we can detect any other disconnect scenarios immediate given the interface of `net.Listener` and `net.Conn`.

(I should note that the code was heavily inspired from https://github.com/golang/net/blob/064579744ee0/netutil/listen.go#L72. The alternative to this change would be to fork `LimitListener` and add a `GaugeFunc` on `len(sem)`, but given it was straightforward to implement this way, it didn't seem necessary.)

bboreham: Ah, I missed you were creating a new object per connection, sorry.

OK, new question: do we know `Close()` is called in all error scenarios?  E.g. if the connection times out.

stevesg: It's possible that a caller does not call `Close` when the underlying connection becomes unusable for some reason, but if that was the case, then the bigger problem is that the `LimitListener` would eventually start blocking connections, as it relies on `Close` being called.

A brief read through the implementation of `http.Server` indicates that it does appear to take a lot of care to `Close()` connections when they have entered a bad state (e.g. handshake fails, write fails), as you would expected.

bboreham (approved):
Thanks!  LGTM

PR #216, opened 2021-05-08
add only-ipv4-support

3Xpl0it3r:
add http server only bind  ipv4 address support

when I used Promtail ,I found Promtail alway bind ipv6, but I want to only bind in ipv4
I debug Promtail,i found Promtail used this library to build an http server.


go.mod
```diff
@@ -25,7 +25,7 @@ require (
 	github.com/sercand/kuberesolver v2.1.0+incompatible
 	github.com/sirupsen/logrus v1.4.2
 	github.com/stretchr/testify v1.4.0
-	github.com/uber/jaeger-client-go v2.22.1+incompatible
+	github.com/uber/jaeger-client-go v2.28.0+incompatible
```
bboreham, line 28: Is this necessary to do ipv4-only?  If yes, please explain; if no, please move to a different PR.

server/server.go
```diff
@@ -158,8 +164,13 @@ func New(cfg Config) (*Server, error) {
 	}, []string{"protocol"})
 	prometheus.MustRegister(tcpConnections)
 
+	// add ipv4 support
+	if cfg.Network == "" {
+		cfg.Network = DefaultNetwork
+	}
```
bboreham, line 170: Can you add to `RegisterFlags()` with the default please, instead of this?

server/server.go
```diff
@@ -44,6 +49,7 @@ type SignalHandler interface {
 // Config for a Server
 type Config struct {
 	MetricsNamespace  string `yaml:"-"`
+	Network           string `yaml:"network"`
```
bboreham, line 52: For consistency this would be `http_listen_network`

bboreham (changes requested):
Thanks for the PR.  The basic idea looks OK, but I don't think you should check in the files in `server/certs` - they are intended to be generated on-the-fly for testing only.

Maybe they should be added to `.gitignore`.

go.mod
3Xpl0it3r: no is not necessary, it looks like go mod replace it

server/server.go
3Xpl0it3r: yeah, sure

server/server.go
3Xpl0it3r: ok, I fix it

3Xpl0it3r:
@bboreham hi ,I reopen a new pr  #217 , this pr I will close it

PR #217, opened 2021-06-15
add tcpv4 support

3Xpl0it3r:
add http/grcp server only bind ipv4 address support


bboreham (approved):
lgtm

pstibrany (approved):
LGTM.

bboreham:
A friend notes this could just as easily be used to bind ipv6-only, for people who prefer that 😉 

Since there is some back-tracking in the commits, I propose to do a squash-and-merge.  If you prefer to do your own re-organisation of commits please go ahead.  I'll wait till tomorrow.

server/server.go
```diff
@@ -31,6 +31,15 @@ import (
 	"github.com/weaveworks/common/signals"
 )
 
+// Dial connects to the address on the named network
+const (
+	// DefaultNetwork  the host resolves to multiple IP addresses,
+	// Dial will try each IP address in order until one succeeds
```
bboreham, line 37: Nit-picking: I would expect Dial to try ipv4 and ipv6 in parallel, per RFC 8305.
But it's irrelevant since we don't call Dial

server/server.go
```diff
@@ -31,6 +31,15 @@ import (
 	"github.com/weaveworks/common/signals"
 )
 
+// Dial connects to the address on the named network
```
bboreham, line 34: This change isn't about Dial but Listen.

bboreham (commented):
Sorry I noticed a comment which is wrong.

server/server.go
3Xpl0it3r: ok ,I fix it

bboreham (approved):
lgtm

PR #218, opened 2021-07-20
tracing: add options to NewFromEnv

hairyhenderson:
I've found myself wanting to set some options (tracer tags, etc), and figured this should work fine without breaking existing code.

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -14,9 +14,13 @@ var (
 )
 
 // installJaeger registers Jaeger as the OpenTracing implementation.
-func installJaeger(serviceName string, cfg *jaegercfg.Configuration) (io.Closer, error) {
+func installJaeger(serviceName string, cfg *jaegercfg.Configuration, options ...jaegercfg.Option) (io.Closer, error) {
 	metricsFactory := jaegerprom.New()
-	closer, err := cfg.InitGlobalTracer(serviceName, jaegercfg.Metrics(metricsFactory))
+
+	// put the metricsFactory earlier so provided options can override it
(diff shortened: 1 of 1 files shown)


bboreham (approved):
lgtm; one point to consider

hairyhenderson:
> one point to consider

Did you miss submitting a comment? I don't see anything 😅

tracing/tracing.go
```diff
@@ -14,9 +14,13 @@ var (
 )
 
 // installJaeger registers Jaeger as the OpenTracing implementation.
-func installJaeger(serviceName string, cfg *jaegercfg.Configuration) (io.Closer, error) {
+func installJaeger(serviceName string, cfg *jaegercfg.Configuration, options ...jaegercfg.Option) (io.Closer, error) {
 	metricsFactory := jaegerprom.New()
-	closer, err := cfg.InitGlobalTracer(serviceName, jaegercfg.Metrics(metricsFactory))
+
+	// put the metricsFactory earlier so provided options can override it
+	opts := append([]jaegercfg.Option{jaegercfg.Metrics(metricsFactory)}, options...)
```
bboreham, line 21: Well, I don't know what happened, but this is what I was going to suggest:
```suggestion
	opts := append([]jaegercfg.Option{}, jaegercfg.Metrics(metricsFactory), options...)
```

It's perhaps a personal preference: I think it makes it more clear the options form one list.

hairyhenderson: @bboreham that doesn't compile 😅

See https://groups.google.com/g/golang-nuts/c/R5RW0oOGSKg for background on why

PR #224, opened 2021-09-22
Request logging is not debug info

ozamosi:
It's voluminous, sure, but it's not just for debugging. Especially things like 4xx
requests.


bboreham:
Closing; please re-open if you come back to it.

PR #226, opened 2021-10-15
Migrating from go-kit/kit/log to the slimmer go-kit/log

dannykopping:
`go-kit` has spun out its logger package, which is all this library uses.


bboreham:
Does that work?  ISTM any time you have `level.Debug` from the two libraries one of them is not going to work.

PR #227, opened 2021-10-17
Speed up debug logging, when filtered out

bboreham:
A benchmark, plus two changes:
 * Filter by log level before anything else - this saves a lot of work walking the stack to find the caller, and a bit more to format the timestamp.  As recommended at https://github.com/go-kit/log/issues/14#issuecomment-945038252
 * Avoid Sprintf when log line is filtered out - defer the Sprintf until after the level check by using an auxilliary struct.

Benchmark results:
```
name      old time/op    new time/op    delta
Debugf-4    2.46µs ± 6%    0.54µs ± 6%  -78.00%  (p=0.008 n=5+5)

name      old alloc/op   new alloc/op   delta
Debugf-4      848B ± 0%      328B ± 0%  -61.32%  (p=0.008 n=5+5)

name      old allocs/op  new allocs/op  delta
Debugf-4      15.0 ± 0%       9.0 ± 0%  -40.00%  (p=0.008 n=5+5)
```

Small improvement in behaviour for `Infoln()`, `Warningln()`, etc.: they no longer add a newline character to the string. Go-kit log lines are intrinsically a single line, with structured fields so it's wrong to add a newline.
Previously this would come out as `msg="My log message\n"` in the log.

### DIFF
--- a/logging/gokit.go
+++ b/logging/gokit.go
@@ -17,6 +17,11 @@ func NewGoKitFormat(l Level, f Format) Interface {
 	} else {
 		logger = log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
 	}
+	return addStandardFields(logger, l)
+}
+
+// stand-alone for test purposes
+func addStandardFields(logger log.Logger, l Level) Interface {
 	logger = level.NewFilter(logger, l.Gokit)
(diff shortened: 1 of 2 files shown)


pracucci (approved):
Smart, good job!

PR #232, opened 2021-11-04
Add a tag with the user-agent to traces

simonswine:
Often it can be useful to understand, which kind of client is behaving in a certain way.

That's why this PR adds a tag to the HTTP trace middleware to annotate the span with the user-agent header of the request.

### DIFF
--- a/middleware/http_tracing.go
+++ b/middleware/http_tracing.go
@@ -29,19 +29,19 @@ func (t Tracer) Wrap(next http.Handler) http.Handler {
 
 			return fmt.Sprintf("HTTP %s - %s", r.Method, op)
 		}),
-		// add a tag with the clients user agent to the span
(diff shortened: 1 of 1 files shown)


bboreham (approved):
Thanks!

Fancy doing remote ip(s) too?

simonswine:
> Fancy doing remote ip(s) too?

I think that is already there (at least for the original client address before load balancer)?

https://github.com/weaveworks/common/blob/05b06253aab8f0667804500deb27d21f8196c27d/middleware/http_tracing.go#L33-L37

bboreham:
Thanks; can we do both inside a single `MWSpanObserver` ?  I imagine that would be more efficient.

simonswine:
@bboreham just updated it accordingly

bboreham (approved):
lgtm, thanks!

Issue #223, opened 2021-09-22
node_exporter/https moved to exporter-toolkit/web


james-callahan:
```
go: finding module for package github.com/prometheus/node_exporter/https
github.com/grafana/agent/cmd/grafana-agent-crow imports
	github.com/weaveworks/common/server imports
	github.com/prometheus/node_exporter/https: module github.com/prometheus/node_exporter@latest found (v1.2.2), but does not contain package github.com/prometheus/node_exporter/https
```

It appears as though `github.com/prometheus/node_exporter/https` was moved to `github.com/prometheus/exporter-toolkit/https` (see https://github.com/prometheus/node_exporter/pull/1907)
That was then moved to `github.com/prometheus/exporter-toolkit/web` (see https://github.com/prometheus/exporter-toolkit/pull/29)

PR #230, opened 2021-10-22
Replace node_exporter https package with exporter-toolkit

rfratto:
The node_exporter/https package has been moved to exporter-toolkit as of v1.1.0. weaveworks/common was still using this package, which was preventing downstream importers (i.e., grafana/agent) from updating their dependency on node_exporter.

This change also required bumping kuberesolver to v2.4.0 (the next version after v2.1.0) which uses a newer version of the gRPC library where type names have changed. The kuberesolver diff between the two versions can be found here: https://github.com/sercand/kuberesolver/compare/v2.1.0...v2.4.0

PR #233, opened 2021-12-10
convert to prometheus/exporter-toolkit

rlankfo:
The `https` package in prometheus/node_exporter has been migrated to prometheus/exporter-toolkit. 

I'm submitting this change to convert over to the new package so that the version of node_exporter can be updated in the Grafana Agent, and other projects that depend on both node_exporter and weaveworks/common.

Fixes https://github.com/weaveworks/common/issues/223

Requires bumping go version to go 1.14


rlankfo:
Closing in favor of #230 (didn't see the other PR before opening this)

PR #234, opened 2021-12-22
middleware: Add gRPC instrumentation utilities from dskit

aknuds1:
Copy gRPC instrumentation utilities from [dskit](https://github.com/grafana/dskit/tree/main/middleware), as suggested offline by @bboreham.


middleware/grpc.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	grpcUtils "github.com/weaveworks/common/grpc"
+	"github.com/weaveworks/common/httpgrpc"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// PrometheusGRPCUnaryInstrumentation records duration of gRPC requests client side.
+func PrometheusGRPCUnaryInstrumentation(metric *prometheus.HistogramVec) grpc.UnaryClientInterceptor {
```
bboreham, line 17: The equivalent server call in this package is `UnaryServerInstrumentInterceptor`, so perhaps `UnaryClientInstrumentInterceptor`?

middleware/grpc.go
```diff
@@ -0,0 +1,104 @@
...
+	return func(ctx context.Context, method string, req, resp interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		start := time.Now()
+		err := invoker(ctx, method, req, resp, cc, opts...)
+		metric.WithLabelValues(method, errorCode(err)).Observe(time.Since(start).Seconds())
+		return err
+	}
+}
+
+// PrometheusGRPCStreamInstrumentation records duration of streaming gRPC requests client side.
+func PrometheusGRPCStreamInstrumentation(metric *prometheus.HistogramVec) grpc.StreamClientInterceptor {
```
bboreham, line 27: The equivalent server call in this package is `StreamServerInstrumentInterceptor`, so perhaps `StreamClientInstrumentInterceptor`?

bboreham (commented):
Bit more context: that code was moved from https://github.com/cortexproject/cortex/blob/ff858460bbb36cbe4708183250f7448242f8d928/pkg/util/middleware/grpc.go
about three months ago, then extended by https://github.com/grafana/dskit/pull/68.

I suggested putting the code here as it is the client equivalent of functions in https://github.com/weaveworks/common/blob/master/middleware/grpc_instrumentation.go.

Given the above, `grpc.go` is not a very explanatory name for the new files.  Maybe `grpc_client_instrumentation.go` ?  Or maybe just put the code in `grpc_instrumentation.go` ?

middleware/grpc.go
aknuds1: @bboreham Thanks, done.

middleware/grpc.go
aknuds1: @bboreham Thanks, done.

aknuds1:
@bboreham Thanks for the review, I renamed the functions and moved them into middleware/grpc_instrumentation.go .

middleware/grpc_instrumentation.go
```diff
@@ -45,3 +47,93 @@ func StreamServerInstrumentInterceptor(hist *prometheus.HistogramVec) grpc.Strea
...
+	if errResp, ok := httpgrpc.HTTPResponseFromError(err); ok {
+		statusFamily := int(errResp.Code / 100)
+		respStatus = strconv.Itoa(statusFamily) + "xx"
+	} else if grpcUtils.IsCanceled(err) {
+		respStatus = "cancel"
+	} else {
+		respStatus = "error"
+	}
+
+	return respStatus
```
pracucci, line 138: [nit] I think we can take the chance to further simplify this, removing `respStatus` at all:

```suggestion
	if err == nil {
		return "2xx"
	}

	if errResp, ok := httpgrpc.HTTPResponseFromError(err); ok {
		statusFamily := int(errResp.Code / 100)
		return strconv.Itoa(statusFamily) + "xx"
	} else if grpcUtils.IsCanceled(err) {
		return "cancel"
	} else {
		return "error"
	}
```

pracucci (approved):
FWIW, LGTM :)

middleware/grpc_instrumentation.go
aknuds1: Applied, thanks.

bboreham (approved):
Sorry, fell off my radar.  LGTM.

PR #236, opened 2022-02-07
Add flag to log requests at info level

colin-stuart:
Adds an optional flag to log requests at the info level, still defaulting to logging requests at the debug level.

### DIFF
--- a/middleware/logging.go
+++ b/middleware/logging.go
@@ -14,10 +14,10 @@ import (
 
 // Log middleware logs http requests
 type Log struct {
-	Log                    logging.Interface
-	LogRequestHeaders      bool // LogRequestHeaders true -> dump http headers at debug log level
-	LogRequestsAtInfoLevel bool // LogRequestAtInfoLevel true -> log requests at info log level
-	SourceIPs              *SourceIPExtractor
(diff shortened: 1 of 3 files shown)


bboreham:
Thanks for the PR. 

It’s now calling `logWithRequest` about six times; can you factor that out?

colin-stuart:
Thanks for the the review @bboreham, let me know what you think about the refactor. Another option would be to add another global convenience function that takes a `level` string and a calls the appropriate `Debugf()`/`Infof()`/etc. function.

bboreham:
What I meant was: call `logWithRequest` once, then hold the resulting `logging.Interface` in a variable for the duration of the `Wrap()` inner function.
It's a relatively expensive operation: https://github.com/weaveworks/common/blob/933588f9873757f6db44841b0919ddaa8c207f98/middleware/logging.go#L23-L38

colin-stuart:
Thanks @bboreham, I see what you meant now!

bboreham:
Looks better now, thanks.  Have you tried it?

colin-stuart:
Yep, enabled it for a service running locally and the requests logs are showing at the info level.

PR #237, opened 2022-02-15
Add user and org labels to observed exemplars

colega:
Exemplars provide information about traces, and traces usually reference to the user that served the request, however, searching the user in the trace is usually a tedious task, and it would be nice to be able to overwiew whether the slow requests (exemplars) from your panel corresponds to the same user.

We could even consider adding exemplars from non-traced requests that have a user ID? This would provide (poor) insights about the request patterns even for the users who run the server without tracing set up.


bboreham (approved):

bboreham:
I agree the caller ID(s) would be useful when there is no trace ID.
However I can also see some downsides:
* current implementation only sends one exemplar per series per scrape, so that change stands the chance of overwriting an exemplar with a trace ID with a different exemplar with no trace ID.
* it might show more overhead.

colega:
Yes, I agree especially on the first point, for people who don't sample all requests that would be a step back.

Can we merge this?

PR #241, opened 2022-05-20
Removed debug log on successfuly gRPC requests from GRPCServerLog

pracucci:
Yesterday I was looking at memory allocations (bytes & objects) in Mimir ingesters with an high volume of small requests, and I've noticed memory allocations triggered by `GRPCServerLog` are non negligible. In this scenario, logging a successful request with debug level (which is then discarded by the filter because we configure logger min level = info) results in about 4% memory allocations:

Screenshot 2022-05-20 at 10 39 17

What's the sentiment if we remove the debug log on successful gRPC requests at all?


pracucci:
Will be superseeded by https://github.com/weaveworks/common/pull/244.

PR #243, opened 2022-06-21
httpgrpc.HTTPResponse from context.Canceled error

colega:
Extends `httpgrpc.HTTPResponseFromError` functionality to map `context.Canceled` errors to 499 status code responses.

I think this makes sense in every situation, and will make all usages much easier, however on the other hand I think this goes beyond the responsibilities of this package.

I'd like to hear opinions on it.

### DIFF
--- a/httpgrpc/httpgrpc.go
+++ b/httpgrpc/httpgrpc.go
@@ -1,6 +1,8 @@
 package httpgrpc
 
 import (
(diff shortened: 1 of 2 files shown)


colega:
Tests already fail with this change, so I'll just close it and let each caller handle their own errors.

PR #244, opened 2022-06-29
Make debug logs on successful requests from GRPCServerLog optional

zenador:
@pracucci asked me to modify https://github.com/weaveworks/common/pull/241 to make the debug log configurable instead of removing it

### DIFF
--- a/middleware/grpc_logging_test.go
+++ b/middleware/grpc_logging_test.go
@@ -13,7 +13,7 @@ import (
 
 func BenchmarkGRPCServerLog_UnaryServerInterceptor_NoError(b *testing.B) {
 	logger := logging.GoKit(level.NewFilter(log.NewNopLogger(), level.AllowError()))
-	l := GRPCServerLog{Log: logger, WithRequest: false}
+	l := GRPCServerLog{Log: logger, WithRequest: false, EnableDebug: false}
 	ctx := context.Background()
 	info := &grpc.UnaryServerInfo{FullMethod: "Test"}
 
--- a/middleware/grpc_logging_test.go
+++ b/middleware/grpc_logging_test.go
@@ -13,7 +13,7 @@ import (
 
 func BenchmarkGRPCServerLog_UnaryServerInterceptor_NoError(b *testing.B) {
 	logger := logging.GoKit(level.NewFilter(log.NewNopLogger(), level.AllowError()))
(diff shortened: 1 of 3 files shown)


middleware/grpc_logging_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"testing"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"google.golang.org/grpc"
+
+	"github.com/weaveworks/common/logging"
+)
+
+func BenchmarkGRPCServerLog_UnaryServerInterceptor_NoError(b *testing.B) {
```
pracucci, line 14: Could you share benchmark results? How to run this benchmark:

1. Run the benchmark both with `master` and this PR code, storing each result to file (I've written the command without testing it, hope I didn't introduce any typo):
   ```
   go test -run '^$' -bench 'BenchmarkGRPCServerLog_UnaryServerInterceptor_NoError' -benchmem ./middleware > before.txt
   ```
2. Use [`benchstat`](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) to compare results:
   ```
   benchstat before.txt after.txt
   ```

pracucci (approved):
LGTM, but I would share benchmark results.

middleware/grpc_logging_test.go
zenador:
```
name                                             old time/op    new time/op    delta
GRPCServerLog_UnaryServerInterceptor_NoError-16     890ns ± 0%      31ns ± 0%   ~     (p=1.000 n=1+1)

name                                             old alloc/op   new alloc/op   delta
GRPCServerLog_UnaryServerInterceptor_NoError-16    1.12kB ± 0%    0.00kB        ~     (p=1.000 n=1+1)

name                                             old allocs/op  new allocs/op  delta
GRPCServerLog_UnaryServerInterceptor_NoError-16      16.0 ± 0%       0.0        ~     (p=1.000 n=1+1)
```

pracucci: Sorry, I forgot one important detail. TO get `benchstat` comparison working, you need to run the benchmark at least twice. Run the test as above, but add `-count=3`

bboreham (approved):
LGTM, thanks @zenador!

PR #245, opened 2022-07-05
server: Removing advanced TLS config parameters - BREAKING CHANGE

aknuds1:
I would like to remove advanced TLS config parameters stemming from github.com/prometheus/exporter-toolkit/web, that were introduced in commit [953ac9fb41437fee0bffcb364333ba624a35e043](https://github.com/weaveworks/common/commit/953ac9fb41437fee0bffcb364333ba624a35e043):

* `cipher_suites`
* `curve_preferences`
* `min_version`
* `max_version`
* `prefer_server_cipher_suites`

The motivation for removing these advanced parameters is that users would most likely not want to change them, and they add corresponding configuration parameters to the Grafana Mimir project, that we don't want. We also think they're not interesting to the Grafana Tempo and Loki projects.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -50,6 +50,14 @@ type SignalHandler interface {
 	Stop()
 }
 
+// TLSConfig contains TLS parameters for Config.
+type TLSConfig struct {
+	TLSCertPath string `yaml:"cert_file"`
+	TLSKeyPath  string `yaml:"key_file"`
+	ClientAuth  string `yaml:"client_auth_type"`
+	ClientCAs   string `yaml:"client_ca_file"`
+}
+
 // Config for a Server
 type Config struct {
 	MetricsNamespace  string `yaml:"-"`
@@ -62,8 +70,8 @@ type Config struct {
 	GRPCListenPort    int    `yaml:"grpc_listen_port"`
 	GRPCConnLimit     int    `yaml:"grpc_listen_conn_limit"`
 
-	HTTPTLSConfig web.TLSStruct `yaml:"http_tls_config"`
-	GRPCTLSConfig web.TLSStruct `yaml:"grpc_tls_config"`
+	HTTPTLSConfig TLSConfig `yaml:"http_tls_config"`
+	GRPCTLSConfig TLSConfig `yaml:"grpc_tls_config"`
 
 	RegisterInstrumentation  bool `yaml:"register_instrumentation"`
 	ExcludeRequestInLog      bool `yaml:"-"`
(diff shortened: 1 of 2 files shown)


bboreham:
This is a breaking change for someone who used this library between merge of #230 and now?

bboreham:
Can you add something to the description what is meant by "hide"?

Here's what I think:
Prior to #230 the `TLSStruct` had just the 4 string settings added by this PR.
#230 changed to a `TLSStruct` with 9 members.
These members are visible if you dump out the YAML for the object, or walk the struct using reflection.
So this PR is to hide them from that.

aknuds1:
> This is a breaking change for someone who used this library between merge of #230 and now?

@bboreham afraid so, in case anyone is using those parameters. Does that imply I should start a changelog, documenting the breaking change? Or do you think this breaking change is simply a bad idea?

bboreham:
Is it possible to set those parameters at all after this change?

If not, I don't think that would be "hide", it would be "remove".
You could hide them from yaml by putting `"-"` as the name; that seems more generous to users.

pracucci:
> You could hide them from yaml by putting "-" as the name; that seems more generous to users.

How can you set these config options using YAML if you use `yaml:"-"`? Also, another discrepancy is that there were no CLI flags registered for the new TLS options, so they were only configurable via YAML. Even so, have you looked at the actual accepted values? There are "list of integers" which I think don't offer a very good UX: I doubt anyone ever used them.

aknuds1:
@bboreham 

> Is it possible to set those parameters at all after this change?

No.

> If not, I don't think that would be "hide", it would be "remove". You could hide them from yaml by putting `"-"` as the name; that seems more generous to users.
That's a good point, didn't think of it.

aknuds1:
> > How can you set these config options using YAML if you use yaml:"-"
> 
> I meant someone using this library as code could then make their own arrangements to set the parameters.

I see @bboreham's point, I think it makes sense.

aknuds1:
@bboreham ok, I can do that.

aknuds1:
Updated PR title and description, plus Git commit message, to clarify the intention and justification.

bboreham (approved):
LGTM; thanks for updating.

bboreham:
> There are "list of integers" which I think don't offer a very good UX

Turns out `prometheus/exporter-toolkit` implements `UnmarshalYAML` so you could say something like:
```
cipher_suites:
- TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
- TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
```

bboreham:
PR to re-introduce some TLS config: #256

PR #246, opened 2022-07-12
fix prometehus client exemplar panic

liguozhong:
https://github.com/grafana/loki/issues/6667


bboreham:
Thanks for this; sorry I didn't get to it sooner.
I merged #247 instead, which is a clean removal; I preferred that over commented-out code.

liguozhong:
> Thanks for this; sorry I didn't get to it sooner. I merged #247 instead, which is a clean removal; I preferred that over commented-out code.

No matter which PR is merged as long as it can solve the problem, this problem really bothered me last week

PR #247, opened 2022-07-19
Revert "Add user and org labels to observed exemplars"

colega:
Reverts weaveworks/common#237

[Prometheus will panic](https://github.com/prometheus/client_golang/blob/3d482bb7cfa10c0cd0d6d3f80cf77b8ce24435bd/prometheus/value.go#L202-L237) if the total length of all labels and their values exceeds 64 runes.
This means that we can't have a user input on the exemplars labels.


colega:
For the context, even though the trace ID comes from the outside, it has a fixed format:

https://github.com/jaegertracing/jaeger-client-go/blob/8d8e8fcfd04de42b8482476abac6a902fca47c18/span_context.go#L366-L395

bboreham (approved):
Thanks

PR #250, opened 2022-08-10
Bugfix server tcp metrics to use the configured registry or the default Prometheus registry

ssncferreira:
Server TCP related metrics `tcp_connections` and `tcp_connections_limit` are currently always registered with the default prometheus registry: https://github.com/prometheus/client_golang/blob/main/prometheus/registry.go#L57
As a result, it can cause an attempt to register duplicate metrics collector failing with a panic. This can happen for instance in a testing scenario.

This PR updates these metrics to use the configured Registry, if any, otherwise, the default Prometheus Registry.
`log` and `gatherer` initializations were also moved for consistency.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -181,19 +181,36 @@ type Server struct {
 
 // New makes a new Server
 func New(cfg Config) (*Server, error) {
+	// If user doesn't supply a logging implementation, by default instantiate
+	// logrus.
+	log := cfg.Log
+	if log == nil {
+		log = logging.NewLogrus(cfg.LogLevel)
+	}
+
+	// If user doesn't supply a registerer/gatherer, use Prometheus' by default.
+	reg := cfg.Registerer
+	if reg == nil {
+		reg = prometheus.DefaultRegisterer
+	}
+	gatherer := cfg.Gatherer
+	if gatherer == nil {
+		gatherer = prometheus.DefaultGatherer
+	}
+
 	tcpConnections := prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: cfg.MetricsNamespace,
 		Name:      "tcp_connections",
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -181,19 +181,36 @@ type Server struct {
 
 // New makes a new Server
 func New(cfg Config) (*Server, error) {
+	// If user doesn't supply a logging implementation, by default instantiate
+	// logrus.
+	log := cfg.Log
+	if log == nil {
+		log = logging.NewLogrus(cfg.LogLevel)
+	}
+
```
bboreham, line 190: Why did this move?

I can guess you moved it for consistency, but you didn't mention this in the description so I have to doubt whether I understand.

bboreham (commented):
Thank you for the PR.  It's fine, but I have some nits:

The commit description is empty.  It's not always easy to refer back to a PR, so please fix that.

The word "Update" in the title seems superfluous.  To me this is a bugfix, or at least making all metrics registered in this function consistent.

And 'log', as noted below.

server/server.go
ssncferreira: Yes, it was simply moved for consistency. Will update the PR description 👍

pracucci (approved):

ssncferreira:
Thank you @bboreham for the review 🙂 

> The commit description is empty. It's not always easy to refer back to a PR, so please fix that.

Updated the commit description with a small description of the change.

> The word "Update" in the title seems superfluous. To me this is a bugfix, or at least making all metrics registered in this function consistent.

Updated the PR title as well as the commit message.

> And 'log', as noted below.

Updated the PR description as well as the commit message.

Let me know if something else is still missing 🙂

bboreham (approved):
LGTM, thanks!

PR #251, opened 2022-08-11
Update the sparse-histograms branch

beorn7:
This pulls in the updates in master as well as the new version of the sparsehistogram of prometheus/client_golang (with the updated protobufs for exposition).


beorn7:
The test failures are expected as prometheus/client_golang needs Go 1.17 (for its sparsehistogram branch).

beorn7:
@bboreham could you have a look? Note that this is only for the experimental sparse-histograms branch.

PR #253, opened 2022-08-18
Fix incorrect import of obsolete github.com/go-kit/kit/log package

hairyhenderson:
This removes a test dependency on `github.com/go-kit/kit/log`, which was otherwise replaced with `github.com/go-kit/log` back in #226

The `go.mod` diff was done by running `go mod tidy` - no changes except for whitespace there

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,37 +3,37 @@ module github.com/weaveworks/common
 go 1.14
 
 require (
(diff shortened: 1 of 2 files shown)


bboreham (approved):
Thanks!

PR #240, opened 2022-05-03
httpgrpc/server: Update NewClient to not use WithBalancerName

metalmatze:
Additionally, gRPC was bumped to v1.34.0 to support the insecure.NewCredentials()

Closes #239 

cc @bboreham

PR #254, opened 2022-08-26
httpgrpc/server: Update NewClient to not use WithBalancerName

### LINKED ISSUE #239: Broken grpc.WithBalancerName
After updating to `google.golang.org/grpc v1.46.0` I cannot compile anymore with the following error:
```
../../../../pkg/mod/github.com/weaveworks/common@v0.0.0-20210913144402-035033b78a78/httpgrpc/server/server.go:137:8: undefined: grpc.WithBalancerName
```

It seems that the `WithBalancerName` was actually removed in grpc after it was deprecated for some time https://github.com/grpc/grpc-go/pull/5232

The code in this repository that needs updating can be found here:
https://github.com/weaveworks/common/blob/f83ccc76d8237acd9d0fa0c68b61298fbb2cb03d/httpgrpc/server/server.go#L137

jpkrohling:
On projects consuming this module as a library, gRPC might be set at
newer versions where the deprecated #WithBalancerName has been removed
already. Given that the version used by this module already offers a
forward-compatible method, this commit uses that instead of the
deprecated #WithBalancerName.

This is a simplified version of #240, without touching the gRPC versions.

Fixes #239

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/httpgrpc/server/server.go
+++ b/httpgrpc/server/server.go
@@ -17,7 +17,6 @@ import (
 	"github.com/sercand/kuberesolver"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/balancer/roundrobin"
 
 	"github.com/weaveworks/common/httpgrpc"
 	"github.com/weaveworks/common/logging"
@@ -132,9 +131,10 @@ func NewClient(address string) (*Client, error) {
 	if err != nil {
 		return nil, err
 	}
+	const grpcServiceConfig = `{"loadBalancingPolicy":"round_robin"}`
 
 	dialOptions := []grpc.DialOption{
-		grpc.WithBalancerName(roundrobin.Name),
+		grpc.WithDefaultServiceConfig(grpcServiceConfig),
 		grpc.WithInsecure(),
 		grpc.WithUnaryInterceptor(grpc_middleware.ChainUnaryClient(
 			otgrpc.OpenTracingClientInterceptor(opentracing.GlobalTracer()),


jpkrohling:
For now, this is just an update of the PR opened by @metalmatze, I'll check the compat with previous versions and mark this PR as ready for review.

jpkrohling:
While I think the newer minor version of gRPC should be compatible with a previous minor in the same major, the change in the sum metrics for the bytes payload is suspicious. Instead of trying to figure out whether a compatibility problem does *not* exist (hard to prove), I decided just to simplify the original PR and just do the minimum necessary changes that work, namely, replacing the `WithBalancerName` with the new alternative, which was already available on the version used by this module.

go.mod
```diff
@@ -3,37 +3,38 @@ module github.com/weaveworks/common
...
-  github.com/stretchr/testify v1.4.0
-  github.com/uber/jaeger-client-go v2.28.0+incompatible
-  github.com/uber/jaeger-lib v2.2.0+incompatible
-  github.com/weaveworks/promrus v1.2.0
-  go.uber.org/atomic v1.5.1 // indirect
-  golang.org/x/net v0.0.0-20210525063256-abc453219eb5
-  golang.org/x/tools v0.0.0-20200825202427-b303f430e36d
-  google.golang.org/grpc v1.31.0
-  gopkg.in/yaml.v2 v2.4.0
+	github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5
```
jpkrohling, line 6: I can revert those, this is mainly a `go mod tidy`. which I ran after seeing the following complaint:

```
middleware/grpc_logging_test.go:7:2: missing go.sum entry for module providing package github.com/go-kit/kit/log (imported by github.com/weaveworks/common/middleware); to add:
```

A diff ignoring the whitespace is here:

```diff
$ git diff -b master go.mod 
diff --git a/go.mod b/go.mod
index dad86ab..8582176 100644
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
        github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
        github.com/davecgh/go-spew v1.1.1
        github.com/felixge/httpsnoop v1.0.1
+       github.com/go-kit/kit v0.9.0
        github.com/go-kit/log v0.1.0
        github.com/gogo/googleapis v1.1.0
        github.com/gogo/protobuf v1.3.0
```

httpgrpc/server/server.go
```diff
@@ -11,13 +11,12 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/grpc-ecosystem/go-grpc-middleware"
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
```
bboreham, line 14: Change with no explanation in description.

bboreham (commented):
Code change looks fine, but can you write a commit description that stands alone, fully describes this change, and doesn't require the reader to chase down three other things to understand it.

bboreham:
`go.mod` was tidied up by #253; please rebase and explain any remaining changes.

go.mod
jpkrohling: Rebasing was enough to prevent changes to the go.mod/sum.

httpgrpc/server/server.go
jpkrohling: It was probably my code editor trying to be smart. I suppose this is because the package name does not match the import path. Anyway, I reverted this change.

jpkrohling:
PR updated. The changes are now minimal.

jpkrohling:
I also just changed the GitHub PR description to match the commit message, as I'm not sure which merge strategy is used in this project.

jpkrohling:
Friendly ping: would this PR be accepted in its current form?

bboreham (approved):
Thanks for the reminder, LGTM.

PR #256, opened 2022-09-01
Allow config of TLS cipher suites and min version

### LINKED ISSUE #248: Enable advanced TLS configuration parameters
Hi @pracucci & @bboreham  Good Day!

We need the full capabilities of TLS config parameters to be available via weaveworks/common package to be configured that are available via exporter-toolkit/web.

We see as part of. #245 it was removed. We are using Cortex and we as per our organization standard we want to use a set of strong ciphers for all the HTTPS listening endpoints.  If we have the above config parameters we can fix it by using the cipher_suites option and prefer_server_cipher_suites.

We see the same problem for Loki, Tempo and Mimir is applicable. Let us know if you need any additional information.

Note: We already enabled the client authentication by setting "RequireAndVerifyClientCert".

bboreham:
Add a single parameter for each, not split across HTTP and gRPC.

Requires change upstream - https://github.com/prometheus/exporter-toolkit/pull/110.  This PR is temporarily using a `replace` directive to use that code.

Downstream projects rely on CLI parameters to generate docstrings, so we add `--server.tls-cipher-suites` and `--server.tls-min-version`. Both CLI and yaml require comma-separated lists of cipher suites, which is different to the yaml array format supported by `prometheus/exporter-toolkit`.

The names accepted are from Go, listed here: https://pkg.go.dev/crypto/tls#pkg-constants

Fixes #248 (partial, but I haven't seen a justification to do the curves, max version, etc).


server/server.go
```diff
@@ -243,15 +247,26 @@ func New(cfg Config) (*Server, error) {
...
-			TLSCertPath: cfg.HTTPTLSConfig.TLSCertPath,
-			TLSKeyPath:  cfg.HTTPTLSConfig.TLSKeyPath,
-			ClientAuth:  cfg.HTTPTLSConfig.ClientAuth,
-			ClientCAs:   cfg.HTTPTLSConfig.ClientCAs,
+			TLSCertPath:  cfg.HTTPTLSConfig.TLSCertPath,
+			TLSKeyPath:   cfg.HTTPTLSConfig.TLSKeyPath,
+			ClientAuth:   cfg.HTTPTLSConfig.ClientAuth,
+			ClientCAs:    cfg.HTTPTLSConfig.ClientCAs,
+			CipherSuites: cipherSuites,
+			MinVersion:   minVersion,
```
aknuds1, line 269: Shouldn't these also be added to `grpcTLSConfig`?

It sounds from the PR description as if they're supposed to be for both HTTP and gRPC:

> Add a single parameter for each, not split across HTTP and gRPC.

aknuds1 (commented):
Wondering if the parameters shouldn't also be applied for gRPC.

go.mod
```diff
@@ -37,3 +37,6 @@ require (
 	google.golang.org/grpc v1.31.0
 	gopkg.in/yaml.v2 v2.4.0
 )
+
+// Export types used in TLS config.  Remove when https://github.com/prometheus/exporter-toolkit/pull/110 is merged.
+replace github.com/prometheus/exporter-toolkit v0.7.0 => github.com/bboreham/exporter-toolkit v0.0.0-20220901124938-52ad245e00a1
```
aknuds1, line 42: The PR is merged, so I guess the `replace` can be dropped.

server/tls_config.go
```diff
@@ -0,0 +1,55 @@
...
+func stringToCipherSuites(s string) ([]web.Cipher, error) {
+	if s == "" {
+		return nil, nil
+	}
+	ciphersSlice := []web.Cipher{}
+	possibleCiphers := allCiphers()
+	for _, cipher := range strings.Split(s, ",") {
+		intValue, ok := possibleCiphers[cipher]
+		if !ok {
+			return nil, fmt.Errorf("Cipher suite %q not recognized", cipher)
```
aknuds1, line 32: [Nit] Conventionally Go errors are not capitalized, right?
```suggestion
			return nil, fmt.Errorf("cipher suite %q not recognized", cipher)
```

server/tls_config_test.go
```diff
@@ -0,0 +1,59 @@
...
+func Test_stringToTLSVersion(t *testing.T) {
+	tests := []struct {
+		name    string
+		arg     string
+		want    web.TLSVersion
+		wantErr bool
+	}{
+		{name: "blank", arg: "", want: 0},
+		{name: "bad", arg: "not-a-version", wantErr: true},
+		{name: "VersionTLS12", arg: "VersionTLS12", want: tls.VersionTLS12},
```
aknuds1, line 46: [Nit] Good looking test, just wondering if it would be good to include all the other supported strings.

aknuds1 (commented):
LGTM, except you should probably drop the `replace` now that the PR is merged.

server/tls_config.go
bboreham: You can get points for reporting / fixing the same thing upstream!
https://github.com/kubernetes/kubernetes/blob/d5f39ebe4d2700eccf094c54b36c5c46302fb4ec/staging/src/k8s.io/component-base/cli/flag/ciphersuites_flag.go#L107

aknuds1: Good point @bboreham :D Taking a look there.

server/tls_config_test.go
bboreham: I can see that I can write a test which checks whether I managed to copy-paste the same strings into the code and the test.  It would not detect if I put the same error in both.
So it doesn't feel valuable.
Maybe I missed something in your suggestion?

aknuds1: It's no biggie, but I think it can be useful to guard against unintended changes in the implementation. Unintended implementation changes can happen over time, in that case it's nice to have an unambiguous test stating the supported inputs. Just a suggestion from my side, feel free to ignore :)

aknuds1 (approved):
LGTM

Edit: Oof,linting fails in CI. An updated dependency probably requires a Go upgrade.

bboreham:
Trying to fix the build at #258

PR #257, opened 2022-09-07
Add badResponseLoggingWriter.Flush

leizor:
This PR adds `Flush()` to the `middleware.badResponseLoggingWriter`, making it implement `http.Flusher`.

I think, ideally, none of this middleware should interfere with what the ultimate HTTP handlers want to try to do. Making a best-effort attempt at `Flush()` seems like the friendliest behavior for middleware; if it can flush, great, otherwise nothing happens.

Additionally, I think this best-effort type behavior plays nicely with [this disclaimer](https://pkg.go.dev/net/http#Flusher) about `http.Flusher`:

> Note that even for ResponseWriters that support Flush, if the client is connected through an HTTP proxy, the buffered data may not reach the client until the response completes.

### DIFF
--- a/middleware/response.go
+++ b/middleware/response.go
@@ -12,9 +12,15 @@ const (
 	maxResponseBodyInLogs = 4096 // At most 4k bytes from response bodies in our logs.
 )
 
-// badResponseLoggingWriter writes the body of "bad" responses (i.e. 5xx
+type badResponseLoggingWriter interface {
+	http.ResponseWriter
+	getStatusCode() int
+	getWriteError() error
+}
+
+// nonFlushingBadResponseLoggingWriter writes the body of "bad" responses (i.e. 5xx
 // responses) to a buffer.
-type badResponseLoggingWriter struct {
+type nonFlushingBadResponseLoggingWriter struct {
 	rw            http.ResponseWriter
 	buffer        io.Writer
 	logBody       bool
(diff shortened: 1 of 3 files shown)


bboreham:
It looks to me like this method will advertise the possibility of `Flush()` on writers that don't support it.
Even though there is a disclaimer, this could break something.  Hyram's Law.

What about this: create a struct embedding `badResponseLoggingWriter` that implements `Flusher`, and select which one to create in `newBadResponseLoggingWriter`.

leizor:
I like that much better - will do!

leizor:
@bboreham I think this is ready for a second look

middleware/response.go
```diff
@@ -67,15 +88,27 @@ func (b *badResponseLoggingWriter) WriteHeader(statusCode int) {
...
+	return b.statusCode
+}
+
+func (b *nonFlushingBadResponseLoggingWriter) getWriteError() error {
+	return b.writeError
+}
+
+func (b *flushingBadResponseLoggingWriter) Flush() {
+	b.f.Flush()
+}
```
pracucci, line 109: [super nit] I would move this below, to keep all `nonFlushingBadResponseLoggingWriter` receiver functions grouped together.

middleware/response_test.go
```diff
@@ -31,12 +31,50 @@ func TestBadResponseLoggingWriter(t *testing.T) {
...
+}
+
+func (rw *nonFlushingResponseWriter) Write(_ []byte) (int, error) {
+	return -1, nil
+}
+
+func (rw *nonFlushingResponseWriter) WriteHeader(_ int) {
+}
+
+func TestNonFlushingBadResponseLoggingWriter(t *testing.T) {
```
pracucci, line 58: [nit] This test both cases, so we could rename it. For example:

```suggestion
func TestBadResponseLoggingWriter_WithAndWithoutFlusher(t *testing.T) {
```

pracucci (approved):
Good job, LGTM!

middleware/response.go
```diff
@@ -23,27 +29,42 @@ type badResponseLoggingWriter struct {
...
-func newBadResponseLoggingWriter(rw http.ResponseWriter, buffer io.Writer) *badResponseLoggingWriter {
-	return &badResponseLoggingWriter{
+// flushingBadResponseLoggingWriter is a badResponseLoggingWriter that
+// implements http.Flusher.
+type flushingBadResponseLoggingWriter struct {
+	nonFlushingBadResponseLoggingWriter
+	f http.Flusher
+}
+
+// newBadResponseLoggingWriter makes a new badResponseLoggingWriter or
```
bboreham, line 39: This comment is slightly incorrect - the type is always `badResponseLoggingWriter` but concretely two other types.  But the concrete type is not part of the observable behaviour so not necessary to mention here.

bboreham (approved):
Code LGTM but small point about a comment.
I'm happy to merge as is, but if you want to address Marco and my comments let me know.

middleware/response.go
leizor: Good point--I'll clean that up!

bboreham (approved):
Thanks!

PR #258, opened 2022-09-09
Update CI build to Go 1.17; drop BUILD_IN_CONTAINER

bboreham:
There is no `weaveworks/build-golang` image newer than Go 1.14, and we don't really need one.
For CI, switch to using CircleCI's image, and for local builds assume the user has Go installed.

Also switch to `golangci-lint`, although I suppressed any pre-existing issues.  Can fix them in a different PR.

The one thing which is tricky to set up locally is protobuf generation, which is not used in CI.
The fact that the three `.pb.go` files in this repo were generated using two different proto commands suggests it was never done consistently.

To avoid leaving protos broken, I have used a 3rd-party image `namely/protoc:1.22_1`, which generates identical
output for two of the protos in this repo.

The third one, httpgrpc, I have not managed to match exactly, but the result is very close.
I have not updated the `.pb.go` files however. Leave that till when someone needs an update.

Also dropped a lot of boilerplate from the makefile which was not needed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 # https://circleci.com/blog/circleci-hacks-reuse-yaml-in-your-circleci-config-with-yaml/
 defaults: &defaults
   docker:
-    - image: weaveworks/build-golang:1.14.4-stretch
+    - image: cimg/go:1.17.13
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM

Thanks @bboreham 🏅

PR #262, opened 2022-11-14
Updated prometheus/exporter-toolkit to v0.8.1

rabenhorst:
Updated prometheus/exporter-toolkit to v0.8.1 to make it work with Prometheus v0.40.1.


go.mod
```diff
@@ -23,17 +23,17 @@ require (
...
 	github.com/uber/jaeger-client-go v2.28.0+incompatible
 	github.com/uber/jaeger-lib v2.2.0+incompatible
 	github.com/weaveworks/promrus v1.2.0
 	go.uber.org/atomic v1.5.1 // indirect
-	golang.org/x/net v0.0.0-20220225172249-27dd8689420f
-	golang.org/x/tools v0.0.0-20200825202427-b303f430e36d
-	google.golang.org/grpc v1.31.0
+	golang.org/x/net v0.0.0-20220909164309-bea034e7d591
+	golang.org/x/tools v0.1.5
+	google.golang.org/grpc v1.47.0
```
bboreham, line 37: Can it be done without this update?  This is an enormous jump.

If not, someone needs to do forwards/backwards compatibility testing to confirm that old and new clients work against old and new servers.

bboreham (commented):
Can we avoid bumping gRPC?
Also, did you `go mod tidy` ?   Seems like a lot of lines added in `go.sum` compared to actual change.

bboreham:
Thanks for the PR, by the way.

bboreham:
Looks like the chain of dependencies that updates gRPC is:

github.com/prometheus/exporter-toolkit@v0.8.1 
-> golang.org/x/oauth2@v0.0.0-20220909003341-f21342109be1 [here](https://github.com/prometheus/exporter-toolkit/blob/v0.8.1/go.mod#L30)
-> cloud.google.com/go/compute@v1.7.0 [here](https://github.com/golang/oauth2/blob/f21342109be1/go.mod#L6)
-> google.golang.org/grpc@v1.47.0 [here](https://github.com/googleapis/google-cloud-go/blob/compute/v1.7.0/go.mod#L17)

I don't see a way to break that, except larger-scale copy-paste.

krajorama:
This is actually needed for native historgrams work as well in mimir-prometheus.

krajorama (commented):
Please run `make protos` and add the result to the PR, thanks

krajorama:
> Please run `make protos` and add the result to the PR, thanks

Actually this lead to errors for Mimir, so let's forget it for now. I'm trying to figure out what happened.

krajorama:
We should merge #264 first.

rabenhorst:
> We should merge #264 first. Fixes the `make protos` issue above.

I rebased on master, ran `make protos` on both macOS arm and linux x86 machine and always get an error:
´´´
gogoproto/gogo.proto: File not found.
httpgrpc/httpgrpc.proto:5:1: Import “gogoproto/gogo.proto” was not found or had errors.
´´´ 

Any idea how to fix this?

cristiangreco:
Drive-by: `prometheus/exporter-toolkit@v0.8.2` has been released today, which fixes https://pkg.go.dev/vuln/GO-2022-1130

krajorama:
> > We should merge #264 first. Fixes the `make protos` issue above.
> 
> I rebased on master, ran `make protos` on both macOS arm and linux x86 machine and always get an error:
> 
> ```
> gogoproto/gogo.proto: File not found.
> httpgrpc/httpgrpc.proto:5:1: Import “gogoproto/gogo.proto” was not found or had errors.
> ```
> 
> Any idea how to fix this?

Need to add a dependency on github.com/gogo/protobuf , at least v1.3.0 and put the path to  gogoproto/gogo.proto in the -I include path of `protoc` compiler.

E.g. in mimir:
https://github.com/grafana/mimir/blob/0f2929359b8b02b8427f1a901839b1805f43d58b/go.mod#L15
https://github.com/grafana/mimir/blob/0f2929359b8b02b8427f1a901839b1805f43d58b/Makefile#L243

And possibly add a tools.go to make `go get` actually download it :( 
https://github.com/weaveworks/common/blob/master/httpgrpc/tools.go

Background:  we found that `make protos` here in weaveworks/common did not reproduce the generated files since it was missing the gogo extension definitions. Now this is added so CI can check protobuf definitions and generated files automatically.

go.mod
krajorama: Actually mimir has been running with 1.45 grpc since 14th June https://github.com/grafana/mimir/pull/2098

krajorama: We've concluded that Mimir's usage of grpc is not proof and I'm going to reuse the server test code (httpgrpc/server_test, with a few additions) to test compatibility of versions:
client using 1.31 grpc -> server 1.47
client 1.47 -> server 1.31

rabenhorst:
> Put the path to gogoproto/gogo.proto in the -I include path of `protoc` compiler. (Since it is already a dependency of thanos-io).

Thx. It worked and there is no change in generated files. Is this expected?

Also I updated exporter-toolkit to v0.8.2.

krajorama:
> > Put the path to gogoproto/gogo.proto in the -I include path of `protoc` compiler. (Since it is already a dependency of thanos-io).
> 
> Thx. It worked and there is no change in generated files. Is this expected?
> 
> Also I updated exporter-toolkit to v0.8.2.

Yes, that's expected, we just wanted to get to a reproducable build where make protos does not result in diffs in git.

go.mod
krajorama: What I did to test gprc message handling compatibility:

Extracted the client and server sides of the `httpgrpc/server/server.go` and httpgrpc/server/server_test.go`  files into https://github.com/krajorama/weaveworks-common-testclient and https://github.com/krajorama/weaveworks-common-testserver .

Added a test to return the OrgId extracted on the server side in the body . And another to test sending a method different then GET.

To run a test I execute the server and copied the URL it listened on, then run the client tests with:
`weaveworks-common-testclient$ SERVER_URL=direct://127.0.0.1:35635 go test -count 1 -v ./...`

The baseline was using the latest weaveworks/common at  v0.0.0-20221125125352-88dcecd133b1, where:
```
require (
        ...
	github.com/weaveworks/common v0.0.0-20221125125352-88dcecd133b1
	google.golang.org/grpc v1.31.0
)
```

Then doing a go.mod replace to get:
```
replace github.com/weaveworks/common v0.0.0-20221125125352-88dcecd133b1 => github.com/rabenhorst/common v0.0.0-20221130173339-6d74945c9349

require (
        ...
	github.com/weaveworks/common v0.0.0-20221125125352-88dcecd133b1
	google.golang.org/grpc v1.47.0
)
```

I've run the test with all combinations: client version x server version. All tests passed.

middleware/grpc_stats_test.go
```diff
@@ -211,7 +211,7 @@ func TestGrpcStatsStreaming(t *testing.T) {
 			received_payload_bytes_bucket{method="gRPC",route="/middleware.EchoServer/Process",le="1.048576e+08"} 5
 			received_payload_bytes_bucket{method="gRPC",route="/middleware.EchoServer/Process",le="2.62144e+08"} 5
 			received_payload_bytes_bucket{method="gRPC",route="/middleware.EchoServer/Process",le="+Inf"} 5
-			received_payload_bytes_sum{method="gRPC",route="/middleware.EchoServer/Process"} 1.5728664e+07
+			received_payload_bytes_sum{method="gRPC",route="/middleware.EchoServer/Process"} 1.5728689e+07
```
bboreham, line 214: Noting this difference suggests to me the code is seeing 5 bytes more on the wire.

bboreham (approved):
As noted in individual comments, this is a big jump in gRPC version, and the stats say something has changed on the wire, but I'm very grateful to @krajorama for testing backwards and forwards compatibility.

middleware/grpc_stats_test.go
krajorama: This turned out to be an artifact from this fix in grpc: https://github.com/grpc/grpc-go/pull/3886 , the old code was reporting wrong wirelength to the statistics callback actually.

PR #264, opened 2022-11-24
Reproducable httpgrpc.pb.go build

krajorama:
**Reproducable httpgrpc.pb.go build**

Use gogo protobuf options explicitly to generate the same Equal, String,
GoString etc functions in the protobuf implementation.

Add tools.go to download the gogo.proto dependency.
Add vendor directory as include for protoc to load the dependency.
Update protoc version otherwise there is a diff in the generated files,
most notably
const _ = proto.GoGoProtoPackageIsVersion3
is replaced with
const _ = proto.GoGoProtoPackageIsVersion2

 **Add check-protos target**
    
This is a simple check to catch if proto generation changes files.

**Check generated protobuf files in CI**

Signed-off-by: György Krajcsovits <gyorgy.krajcsovits@grafana.com>


Makefile
```diff
@@ -11,17 +11,17 @@ RM := --rm
 GO_FLAGS := -ldflags "-extldflags \"-static\" -linkmode=external -s -w" -tags netgo -i
 
 # A 3-year-old image which is a reasonable match for the last time the files were generated.
-PROTOC_IMAGE=namely/protoc:1.22_1
+PROTOC_IMAGE=namely/protoc:1.23_0
```
bboreham, line 14: Can you mention somewhere why this changed?

bboreham (commented):
Great work!  One small request, maybe it can go in a commit message.

Makefile
krajorama: There's 3 small(?) changes in the generated files otherwise, I've noted the most interesting one in the commit message and above in the PR description

bboreham (approved):
LGTM, thanks.

PR #265, opened 2022-11-25
Make gogo.proto import vendoring agnostic

krajorama:
prometheus/alertmanager imports gogo.proto as gogoproto/gogo.proto

This is because it does not use vendoring and needs to look up gogoproto in the module cache, which results in a path like
/home/user/go/pkg/mod/github.com/gogo/protobuf@v1.3.2.

To avoid having @v1.3.2 in the proto definition it then sets this as an include path and import only gogoproto/gogo.proto.

This is an issue because protoc is not smart enough to recognize that github.com/gogo/protobuf/gogoproto/gogo.proto and gogoproto/gogo.proto are the same.

Signed-off-by: György Krajcsovits <gyorgy.krajcsovits@grafana.com>


krajorama:
Testing mimir build in https://github.com/grafana/mimir/pull/3524

PR #269, opened 2022-12-01
Small testcode fixes

krajorama:


httpgrpc/server/server.go
```diff
@@ -161,7 +161,7 @@ func HTTPRequest(r *http.Request) (*httpgrpc.HTTPRequest, error) {
 	}
 	return &httpgrpc.HTTPRequest{
 		Method:  r.Method,
-		Url:     r.RequestURI,
+		Url:     r.URL.Path,
```
bboreham, line 164: This is not obviously test code; it's not in a file named "test" or commented as such.

Maybe it's a valid change, but would need to be flagged as a breaking change.

krajorama: fair enough, I'll remove it from the diff

bboreham (approved):
Thanks!

PR #273, opened 2023-01-17
Use ChainInterceptor from upstream grpc

bboreham:
And eliminate dependency on `grpc-ecosystem/go-grpc-middleware`. Functionality is identical; the feature was added upstream in v1.21 for client side and v1.28 for server side.


pstibrany (approved):
Seems like no-brainer to me.

PR #277, opened 2023-02-07
Log middleware now accepts a list of headers to exclude

jesusvazquez:
Previously the log middleware would exclude printing only the headers "Cookie", "Authorization" and "X-Csrf-Token". I believe the users of this library may want to ommit more headers than those.

To do so I've changed how the middleware is initialize to optionally add more headers to the list.

Also I've added a unit test to make sure the behavior works as expected.

Also I've added an optional configuration parameter to the server to set LogRequestHeaders from there. This parameter was already an option in the log middleware but was not exposed in the server.

Signed-off-by: Jesus Vazquez <jesus.vazquez@grafana.com>

### DIFF
--- a/middleware/logging.go
+++ b/middleware/logging.go
@@ -18,6 +18,31 @@ type Log struct {
 	LogRequestHeaders     bool // LogRequestHeaders true -> dump http headers at debug log level
 	LogRequestAtInfoLevel bool // LogRequestAtInfoLevel true -> log requests at info log level
 	SourceIPs             *SourceIPExtractor
+	HttpHeadersToExclude  map[string]bool
+}
+
+var defaultExcludedHeaders = map[string]bool{
+	"Cookie":        true,
+	"X-Csrf-Token":  true,
+	"Authorization": true,
+}
+
+func NewLogMiddleware(log logging.Interface, logRequestHeaders bool, logRequestAtInfoLevel bool, sourceIPs *SourceIPExtractor, headersList []string) Log {
+	httpHeadersToExclude := map[string]bool{}
+	for header := range defaultExcludedHeaders {
+		httpHeadersToExclude[header] = true
+	}
+	for _, header := range headersList {
+		httpHeadersToExclude[header] = true
+	}
+
+	return Log{
+		Log:                   log,
+		LogRequestHeaders:     logRequestHeaders,
+		LogRequestAtInfoLevel: logRequestAtInfoLevel,
+		SourceIPs:             sourceIPs,
+		HttpHeadersToExclude:  httpHeadersToExclude,
+	}
 }
 
 // logWithRequest information from the request and context as fields.
@@ -45,7 +70,7 @@ func (l Log) Wrap(next http.Handler) http.Handler {
 		uri := r.RequestURI // capture the URI before running next, as it may get rewritten
 		requestLog := l.logWithRequest(r)
 		// Log headers before running 'next' in case other interceptors change the data.
-		headers, err := dumpRequest(r)
+		headers, err := dumpRequest(r, l.HttpHeadersToExclude)
 		if err != nil {
 			headers = nil
 			requestLog.Errorf("Could not dump request headers: %v", err)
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -162,7 +165,9 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&cfg.LogSourceIPs, "server.log-source-ips-enabled", false, "Optionally log the source IPs.")
 	f.StringVar(&cfg.LogSourceIPsHeader, "server.log-source-ips-header", "", "Header field storing the source IPs. Only used if server.log-source-ips-enabled is true. If not set the default Forwarded, X-Real-IP and X-Forwarded-For headers are used")
 	f.StringVar(&cfg.LogSourceIPsRegex, "server.log-source-ips-regex", "", "Regex for matching the source IPs. Only used if server.log-source-ips-enabled is true. If not set the default Forwarded, X-Real-IP and X-Forwarded-For headers are used")
+	f.BoolVar(&cfg.LogRequestHeaders, "server.log-request-headers-enabled", false, "Optionally log request headers.")
```
bboreham, line 168: Right now I feel we do not want to add this flag for all `Server`s, and that it was deliberately omitted before.
Consumers of the library who do want to expose this config to their users can do it themselves and pass `Log`.

(I do not appear to have thought about this for `LogSourceIPsHeader`, etc., but here we are)

server/server.go
```diff
@@ -162,7 +165,9 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
...
 	f.BoolVar(&cfg.LogRequestAtInfoLevel, "server.log-request-at-info-level-enabled", false, "Optionally log requests at info level instead of debug level.")
+	f.StringVar(&cfg.LogRequestExcludeHeadersList, "server.log-request-exclude-headers-list", "", "Comma separated list of headers to be excluded from the request log. By default headers excluded are Authorization, Cookie and X-Csrf-Token. Other headers specified here will just add to the list.")
```
bboreham, line 170: As above, I don't think we want every program using this package to add this verbiage.

middleware/logging_test.go
```diff
@@ -96,6 +97,64 @@ func TestLoggingRequestsAtInfoLevel(t *testing.T) {
...
+			}
+			recorder := httptest.NewRecorder()
+
+			w := errorWriter{
+				err: nil,
+				w:   recorder,
+			}
+			loggingHandler.ServeHTTP(w, req)
+
+			body := buf.String()
```
bboreham, line 150: "body" seems a confusing choice of name for this, which is the log output, where we are dealing with http requests which have a member `Body`.

middleware/logging_test.go
```diff
@@ -96,6 +97,64 @@ func TestLoggingRequestsAtInfoLevel(t *testing.T) {
...
+			}
+			loggingHandler := loggingMiddleware.Wrap(http.HandlerFunc(handler))
+
+			req := httptest.NewRequest("GET", "http://example.com/foo", nil)
+			for _, header := range tc.setHeaderList {
+				req.Header.Set(header, header)
+			}
+			recorder := httptest.NewRecorder()
+
+			w := errorWriter{
```
bboreham, line 144: I don't think this `errorWriter` is doing anything, with its `err` member always nil.

bboreham (commented):
Thanks for the PR.  Some thoughts below.

middleware/logging.go
```diff
@@ -95,15 +113,11 @@ var Logging = Log{
...
-func dumpRequest(req *http.Request) ([]byte, error) {
+func dumpRequest(req *http.Request, httpHeadersToExclude map[string]bool) ([]byte, error) {
 	var b bytes.Buffer
 
 	// Exclude some headers for security, or just that we don't need them when debugging
-	err := req.Header.WriteSubset(&b, map[string]bool{
-		"Cookie":        true,
-		"X-Csrf-Token":  true,
-		"Authorization": true,
-	})
```
fionaliao, line 106: The `middleware.Log` struct is exported so users might be initialising it directly rather than just using the default instance created by the server (e.g. if they wanted to modify a config that's not exposed as a flag). In that case, if we merge this change, these users would have `HttpHeadersToExclude` set to an empty slice and the default headers will be logged. I think this library is widely used enough that we need to be careful when making changes like that. Therefore we should always remove the default headers, even if they're not in the `HttpHeadersToExclude` slice. 

@bboreham Do you think this would be an issue in reality?

jesusvazquez: In case this is an issue I've created this commit https://github.com/weaveworks/common/pull/277/commits/65d65649d8d366de8cfd801db246d0d081a230fb

bboreham: @fionaliao yes, this is a good point. Thanks!

bboreham (approved):
lgtm

PR #280, opened 2023-03-14
Enable Prometheus native histograms for request_duration_seconds

beorn7:
This is an opt-in via setting Config.MetricsNativeHistogramFactor so that nobody will expose native histograms by accident. Even if native histograms are enabled, the conventional histograms are still maintained and presented as before to scrapers incapabale of scraping native histograms.

Native histograms were tested previously in the sparse-histograms branch without any problems. This commit obsoletes the sparse-histograms branch.

The histograms request_message_bytes and response_message_bytes are left as conventional histograms for now. Maybe they would work well as native histograms with a bucket factor of 2, as high resolution is not the aim here. But we can decide about that once we have gathered more experience with the native histograms for request_duration_seconds.

The settings for limiting the bucket count
(i.e. NativeHistogramMaxBucketNumber = 100 and
NativeHistogramMinResetDuration = 1h) are both conservative and most likely good enough for the expected use cases. We can make them configurable later, if users feel the need. The idea here is to not add too many knobs prematurely and thereby avoid confusion.

Also includes an update of prometheus/client_golang to v1.14.0 as that is the earliest release supporting native histograms.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/opentracing/opentracing-go v1.1.0
 	github.com/pkg/errors v0.9.1
 	github.com/pmezard/go-difflib v1.0.0
-	github.com/prometheus/client_golang v1.13.0
+	github.com/prometheus/client_golang v1.14.0
 	github.com/prometheus/exporter-toolkit v0.8.2
 	github.com/sercand/kuberesolver v2.4.0+incompatible
 	github.com/sirupsen/logrus v1.6.0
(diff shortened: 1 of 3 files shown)


beorn7:
@bboreham a number of people want to try out native histograms in production. With the change in this PR, they don't need to build from a branch anymore. For those that don't use native histograms, nothing should change.

bboreham (commented):
Thanks for this.  Can you mention in description the upgrade of client_golang to 1.14, which seems to be the first release supporting native histograms?

I feel the inline comment is a bit long considering this is not a metrics package - could it point to more detailed docs?

beorn7:
Feedback addressed. PHAL.

bboreham (approved):
lgtm, thanks

PR #283, opened 2023-05-04
Add Unwrap method to http.ResponseWriter implementations.

pstibrany:
This PR adds `Unwrap` method to all implementations of `http.ResponseWriter`.

New https://pkg.go.dev/net/http#ResponseController introduced in Go 1.20 uses `Unwrap` method to get access to original `http.ResponseWriter` or underlying connection.

My use case: I'd like to use `http.ResponseController` to set longer timeouts on HTTP endpoints for uploading big files, overriding global `-server.http-write-timeout` and `-server.http-read-timeout`. 

This PR also updates httpsnoop library to version 1.0.3, which supports `Unwrap` method as well (since 1.0.2) (https://github.com/felixge/httpsnoop/issues/8).

(To make weaveworks/common fully compatible with new `http.ResponseController`, `opentracing-contrib/go-stdlib` library needs an update too, after https://github.com/opentracing-contrib/go-stdlib/pull/68 is merged).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/aws/aws-sdk-go v1.27.0
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
 	github.com/davecgh/go-spew v1.1.1
-	github.com/felixge/httpsnoop v1.0.1
+	github.com/felixge/httpsnoop v1.0.3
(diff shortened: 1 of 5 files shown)


bboreham (approved):
Thanks, LGTM.

Musing: I wonder if some of the other methods on the interceptors could be dropped, because Go will now call them via `Unwrap()`. But not now since I want this library to be usable by people on non-latest Go version.

PR #284, opened 2023-05-04
Make DisableRequestSuccessLog configurable.

DylanGuedes:
- Make `DisableRequestSuccessLog` configurable by adding a YAML key to its definition
- Modify the http logging middleware to not log successful requests if `DisableRequestSuccessLog` is truthy. This is especially useful to avoid excessive logging for servers that receive too many successful requests without forcing the logger to change from debug level.

### DIFF
--- a/middleware/logging.go
+++ b/middleware/logging.go
@@ -14,11 +14,12 @@ import (
 
 // Log middleware logs http requests
 type Log struct {
-	Log                   logging.Interface
-	LogRequestHeaders     bool // LogRequestHeaders true -> dump http headers at debug log level
-	LogRequestAtInfoLevel bool // LogRequestAtInfoLevel true -> log requests at info log level
-	SourceIPs             *SourceIPExtractor
-	HttpHeadersToExclude  map[string]bool
+	Log                      logging.Interface
+	DisableRequestSuccessLog bool
+	LogRequestHeaders        bool // LogRequestHeaders true -> dump http headers at debug log level
+	LogRequestAtInfoLevel    bool // LogRequestAtInfoLevel true -> log requests at info log level
+	SourceIPs                *SourceIPExtractor
+	HttpHeadersToExclude     map[string]bool
 }
 
 var defaultExcludedHeaders = map[string]bool{
@@ -27,7 +28,7 @@ var defaultExcludedHeaders = map[string]bool{
 	"Authorization": true,
 }
 
-func NewLogMiddleware(log logging.Interface, logRequestHeaders bool, logRequestAtInfoLevel bool, sourceIPs *SourceIPExtractor, headersList []string) Log {
+func NewLogMiddleware(log logging.Interface, logRequestHeaders bool, logRequestAtInfoLevel bool, sourceIPs *SourceIPExtractor, headersList []string, disableSucessLog bool) Log {
 	httpHeadersToExclude := map[string]bool{}
 	for header := range defaultExcludedHeaders {
 		httpHeadersToExclude[header] = true
@@ -37,11 +38,12 @@ func NewLogMiddleware(log logging.Interface, logRequestHeaders bool, logRequestA
 	}
 
 	return Log{
-		Log:                   log,
-		LogRequestHeaders:     logRequestHeaders,
-		LogRequestAtInfoLevel: logRequestAtInfoLevel,
-		SourceIPs:             sourceIPs,
-		HttpHeadersToExclude:  httpHeadersToExclude,
+		Log:                      log,
+		LogRequestHeaders:        logRequestHeaders,
+		LogRequestAtInfoLevel:    logRequestAtInfoLevel,
+		SourceIPs:                sourceIPs,
+		HttpHeadersToExclude:     httpHeadersToExclude,
+		DisableRequestSuccessLog: disableSucessLog,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


middleware/logging.go
```diff
@@ -94,6 +96,11 @@ func (l Log) Wrap(next http.Handler) http.Handler {
 
 			return
 		}
+
+		if statusCode >= 200 && statusCode < 300 && l.DisableRequestSuccessLog {
```
bboreham, line 100: Would you consider phrasing the three possibilities as `switch` cases?

middleware/logging.go
```diff
@@ -14,11 +14,12 @@ import (
 
 // Log middleware logs http requests
 type Log struct {
-	Log                   logging.Interface
-	LogRequestHeaders     bool // LogRequestHeaders true -> dump http headers at debug log level
-	LogRequestAtInfoLevel bool // LogRequestAtInfoLevel true -> log requests at info log level
-	SourceIPs             *SourceIPExtractor
-	HttpHeadersToExclude  map[string]bool
+	Log                      logging.Interface
+	DisableRequestSuccessLog bool
```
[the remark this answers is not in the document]
DylanGuedes: Good point. I can definitely rewrite it but I did it this way because:
- It is more consistent with the existing DisableRequestSuccessLog
- It allow `weaveworks/common` users to leave it unset by default (i.e: won't introduce an unwanted change)
That said, I'm happy with either; based on those two points, do you prefer `DisableRequestSuccessLog` (which is false by default) ot `LogSuccesfulRequests` (which is truthy by default)? If you prefer the former, would you like to see a rewrite of  `DisableRequestSuccessLog` too?

middleware/logging.go
DylanGuedes: sure thing.

server/server.go
```diff
@@ -83,7 +83,7 @@ type Config struct {
 
 	RegisterInstrumentation  bool `yaml:"register_instrumentation"`
 	ExcludeRequestInLog      bool `yaml:"-"`
-	DisableRequestSuccessLog bool `yaml:"-"`
+	DisableRequestSuccessLog bool `yaml:"disable_request_success_log"`
```
bboreham, line 86: Somehow my comment here got lost. 
I think if you do this without adding a flag it shows up in docs with no explanation, which is undesirable.
[Example](https://github.com/grafana/mimir/pull/4922/files#diff-809450d2d139bdb2005dd008d79c44bbad79aa5f76dacd68a5b46c7f3fe669c6R573)
I would prefer you expose in your own program, and leave it unexposed in weaveworks/common, as ExcludeRequestInLog is.

DylanGuedes: aháaaa I definitely overlooked that, great call. That for sure works, thank you.

middleware/logging.go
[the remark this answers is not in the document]
bboreham: > It is more consistent with the existing DisableRequestSuccessLog

Can you rephrase?  It looks to me like you are adding DisableRequestSuccessLog.

> leave it unset by default (i.e: won't introduce an unwanted change)

This doesn't apply to anyone calling `NewLogMiddleware`; do you think anyone is doing it another way?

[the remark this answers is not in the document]
DylanGuedes: > > It is more consistent with the existing DisableRequestSuccessLog
> 
> Can you rephrase? It looks to me like you are adding DisableRequestSuccessLog.

I was referring to the existing `DisableRequestSuccessLog` defined under `server/server.go`.

> > leave it unset by default (i.e: won't introduce an unwanted change)
> 
> This doesn't apply to anyone calling `NewLogMiddleware`; do you think anyone is doing it another way?

Hmm I thought that mostly users were only doing `server.New()`, without defining `NewLogMiddleware` directly. I'll check how we do it, but tbh your suggestion of defining a flag on my program will make this work any way I prefer, so indeed I shouldn't be too concerned about it.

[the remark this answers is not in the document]
bboreham: OK, I didn't know there was already one in `middleware.GRPCServerLog`.
In that case, you can add one for `middleware.Log` and make it configurable in exactly the same way, no extra parameters.

server/server.go
bboreham: Now I understand the context a bit better, my preference is to leave this as `"-"`.
You can add any config or flags you want in your downstream application.

DylanGuedes: Addressed by https://github.com/weaveworks/common/pull/284/commits/ed6819b7603030a18fc24bec7dbfbce39f318a05.

middleware/logging.go
DylanGuedes: Addressed by https://github.com/weaveworks/common/pull/284/commits/565676c80fb17b86d151a6863c6f18517ea0c809. WDYT?

middleware/logging.go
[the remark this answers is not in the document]
DylanGuedes: I'm not sure if that's what you had in mind but I pushed https://github.com/weaveworks/common/pull/284/commits/d9330659bfde5eaaa4e1877f3e56a39ed2dc2648 that does the following:
- Remove `DisableRequestSuccessLog` from the list of parameters, to avoid a breaking change to `NewLogMiddleware` users
- Make `LogMiddleware.DisableRequestSuccessLog` default value to be whatever is set on `server.Config.DisableRequestSuccessLog`.

middleware/logging_test.go
```diff
@@ -57,6 +58,51 @@ func TestBadWriteLogging(t *testing.T) {
...
+		recorder := httptest.NewRecorder()
+
+		w := errorWriter{
+			err: tc.err,
+			w:   recorder,
+		}
+		loggingHandler.ServeHTTP(w, req)
+		content := buf.String()
+
+		require.True(t, strings.Contains(content, tc.logContains))
```
bboreham, line 102: This is a bit weak for the second test case.
Wouldn't it be better to require that it does _not_ contain the request?

bboreham (approved):
Thanks.  I have one more suggestion.

middleware/logging_test.go
DylanGuedes: Done! WDYT?

Issue #285, opened 2023-05-05
Restarting a server leads to metrics re-registration issues


thampiotr:
### Problem:
When I run a `common/server/server.go`, then shut it down and start a new one (e.g. with a different port) in the same metrics namespace, I get a panic due to re-registration of metrics. 

I tried working around this by keeping track of the metrics and explicitly `Unregister`-ing them, but this will still lead to errors such as:
```
An error has occurred while serving metrics:

1 error(s) occurred:
* collected metric "my_metric_name" { label:<name:"test_label" value:"test_value" > gauge:<value:0 > } was collected before with the same name and label values
```

This is because even after unregistering, it is not possible to register a new `Collector` that is inconsistent with the unregistered collector (as per [`prometheus.Registerer` docs](https://github.com/prometheus/client_golang/blob/main/prometheus/registry.go#L133)).

### Proposed solution:
Instead of using `reg.MustRegister(<collector>)`, the server should use `reg.Register`. When the `AlreadyRegisteredError` is returned by `reg.Register`, it will contain the previously registered `Collector`. The server should use that previously registered `Collector` instead. 

### Discussion
I’m curious to hear what are your thoughts on this approach? 
Are there any alternatives that you would recommend? 
Would you be open to a contribution?

bboreham:
I think, under your proposal, if this library was used in a program which happened to also use the same metric name for something else, things could get confusing.

How about splitting `server.New()` into a function which registers the metrics and a function which creates the server?  And leave the existing API creating a singleton which lasts the life of the program.

thampiotr:
> I think, under your proposal, if this library was used in a program which happened to also use the same metric name for something else, things could get confusing.

Thanks @bboreham, that's a very good point!

> How about splitting `server.New()` into a function which registers the metrics and a function which creates the server? And leave the existing API creating a singleton which lasts the life of the program.

I've drafted this solution to see how it looks here: https://github.com/weaveworks/common/pull/289. It looks good to me - let me know if you have any early feedback. I'll add some tests before marking it as ready for review.

thampiotr:
I've added tests and the change is now ready for review: https://github.com/weaveworks/common/pull/289

PR #289, opened 2023-05-25
Allow users to manually register metrics

thampiotr:
As discussed in https://github.com/weaveworks/common/issues/285 - this PR adds ability to manually register metrics, for example, when users restart the Server, but want to reuse existing and already registered metrics.

It does so by introducing a new, alternative constructor function for the server so a user can pick the preferred implementation:

```
// New makes a new Server. It will panic if the metrics cannot be registered.
func New(cfg Config) (*Server, error) {
	metrics := NewServerMetrics(cfg)
	metrics.MustRegister(cfg.registererOrDefault())
	return newServer(cfg, metrics)
}

// NewWithMetrics makes a new Server using the provided Metrics. It will not attempt to register the metrics,
// the user is responsible for doing so.
func NewWithMetrics(cfg Config, metrics *Metrics) (*Server, error) {
	return newServer(cfg, metrics)
}
```

Fixes https://github.com/weaveworks/common/issues/285

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -7,8 +7,8 @@ import (
 	"errors"
 	"flag"
 	"io"
-	"io/ioutil"
 	"net/http"
+	"os"
 	"os/exec"
 	"strconv"
 	"testing"
@@ -20,7 +20,7 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
 
-	google_protobuf "github.com/golang/protobuf/ptypes/empty"
+	protobuf "github.com/golang/protobuf/ptypes/empty"
 	"github.com/gorilla/mux"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


server/server_test.go
```diff
@@ -578,7 +594,7 @@ func TestTLSServer(t *testing.T) {
 
 	require.Equal(t, res.StatusCode, http.StatusOK)
 
-	body, err := ioutil.ReadAll(res.Body)
+	body, err := io.ReadAll(res.Body)
```
bboreham, line 606: Unrelated; belongs in linter commit?

server/server_test.go
```diff
@@ -435,7 +439,19 @@ func TestHTTPInstrumentationMetrics(t *testing.T) {
 		# TYPE tcp_connections gauge
 		tcp_connections{protocol="http"} 0
 		tcp_connections{protocol="grpc"} 0
-	`), "request_message_bytes", "response_message_bytes", "inflight_requests", "tcp_connections"))
+		
+		# HELP tcp_connections_limit The max number of TCP connections that can be accepted (0 means no limit).
```
bboreham, line 451: I'm unclear why this changed.  If it's unrelated, please put it back.

server/server_test.go
```diff
@@ -126,14 +129,14 @@ func TestDefaultAddresses(t *testing.T) {
 		w.WriteHeader(204)
 	})
 
-	go server.Run()
+	go func() { require.NoError(t, server.Run()) }()
```
bboreham, line 132: I don't think require.NoError will work in a background goroutine.
https://pkg.go.dev/testing#T.FailNow

(there are multiple instances of this in the PR)

server/server_test.go
```diff
@@ -105,6 +105,9 @@ func TestTCPv4Network(t *testing.T) {
 
 		require.NoError(t, srv.grpcListener.Close())
 		require.NotNil(t, <-errChan)
+
+		// So that address is freed for further tests.
+		srv.Shutdown()
```
bboreham, line 110: Is this a linter-driven change?

server/server_test.go
```diff
@@ -145,11 +148,16 @@ func TestDefaultAddresses(t *testing.T) {
 }
 
 func TestErrorInstrumentationMiddleware(t *testing.T) {
+	reg := prometheus.NewRegistry()
+	prometheus.DefaultRegisterer = reg
+	prometheus.DefaultGatherer = reg
+
```
bboreham, line 154: Please explain this change somewhere.  Is it wise to change such global variables?
I don't think it is linter-driven as claimed by the commit description.

bboreham (commented):
Code change looks fine; I have a few points in the tests.

server/server_test.go
thampiotr: It expanded the scope of the test, but it is indeed not related. I've put it back to previous version.

server/server_test.go
thampiotr: No - I was running tests multiple times with `go test github.com/weaveworks/common/server -count 10` and ran into few issues. I've reverted this now, so that linter commit only contains linter changes.

bboreham (approved):
LGTM, thanks!

PR #292, opened 2023-07-05
grpc: use errors.Is to check if error is Canceled

bboreham:
This improves behaviour when one error wraps another.


pstibrany (approved):

bboreham:
I'm going to merge this even though CI is broken; I'll check it out in my larger project and fix up if it's wrong.

PR #293, opened 2023-07-13
middleware: add a way to stop errors being logged

bboreham:
E.g. if the error is caused by overload, then we don't want to log it because that uses more resource.

### DIFF
--- a/middleware/logging.go
+++ b/middleware/logging.go
@@ -51,7 +51,7 @@ func NewLogMiddleware(log logging.Interface, logRequestHeaders bool, logRequestA
 type DoNotLogError struct{ Err error }
 
 func (i DoNotLogError) Error() string        { return i.Err.Error() }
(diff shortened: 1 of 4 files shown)


bboreham:
Moved to draft because I would like to get more agreement how this should be implemented.
https://github.com/grafana/mimir/pull/5494#pullrequestreview-1528283384

middleware/logging.go
```diff
@@ -83,6 +91,9 @@ func (l Log) Wrap(next http.Handler) http.Handler {
 		statusCode, writeErr := wrapped.getStatusCode(), wrapped.getWriteError()
 
 		if writeErr != nil {
+			if errors.Is(writeErr, DoNotLogError{}) {
```
bboreham, line 94: This isn't right: the `writeError` is an error from `Write()`, not the underlying operation error.

bboreham:
Replaced by #299

PR #294, opened 2023-07-17
Add ability to pass TLS certs and keys inline

fayzal-g:
With https://github.com/prometheus/exporter-toolkit/pull/158, it is now possible pass forward TLS certificates and keys inline as part of the `TLSConfig` for the Server.

This PR also extends the `TLSConfig` within `Server.go` to allow the TLS certificate, private key and client CAs to be passed inline as a string. The content of this string is expected to be exactly the as a file, a PEM-encoded sequence of bytes. All parameters are then passed forward to `web.TLSConfig` within `prometheus/exporter-toolkit` as before.


server/server.go
```diff
@@ -52,10 +53,13 @@ type SignalHandler interface {
 
 // TLSConfig contains TLS parameters for Config.
 type TLSConfig struct {
-	TLSCertPath string `yaml:"cert_file"`
-	TLSKeyPath  string `yaml:"key_file"`
-	ClientAuth  string `yaml:"client_auth_type"`
-	ClientCAs   string `yaml:"client_ca_file"`
+	TLSCert       string `yaml:"cert"`
+	TLSKey        string `yaml:"key"`
```
bboreham: Would we want to make this a `config.Secret` to avoid printing the value ?

fayzal-g: Good point. Updated.

PR #295, opened 2023-07-21
Fix issue where successful streaming requests are not always captured in metrics

charleskorn:
According to https://pkg.go.dev/google.golang.org/grpc#ClientConn.NewStream, it's safe to not call `Recv()` on a `grpc.ClientStream` until it's exhausted if you clean up the stream in other ways:

> To ensure resources are not leaked due to the stream returned, one of the following actions must be performed:
> 
> 1. Call Close on the ClientConn.
> 2. Cancel the context provided.
> 3. Call RecvMsg until a non-nil error is returned. A protobuf-generated client-streaming RPC, for instance, might use the helper function CloseAndRecv (note that CloseSend does not Recv, therefore is not guaranteed to release all resources).
> 4. Receive a non-nil, non-io.EOF error from Header or SendMsg.

However, the current implementation of `instrumentedClientStream` does not correctly handle choices 1 and 2: if the `ClientConn` is closed and no further action is taken with the stream, or if the context provided is cancelled, metrics for the request are not recorded.

This PR addresses this issue for choice 2 (context cancellation). The fix is inspired by `opentracing-contrib/go-grpc`'s equivalent implementation: [source](https://github.com/opentracing-contrib/go-grpc/blob/master/client.go)

Fixing the issue for choice 1 (calling `Close` on `ClientConn`) is more involved, so in the interests of keeping this PR small, I've chosen to focus on just the context cancellation case here.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,15 +23,16 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/prometheus/client_golang v1.14.0
+	github.com/prometheus/client_model v0.3.0
 	github.com/prometheus/exporter-toolkit v0.8.2
 	github.com/sercand/kuberesolver/v4 v4.0.0
 	github.com/sirupsen/logrus v1.6.0
-	github.com/soheilhy/cmux v0.1.5 // indirect
+	github.com/soheilhy/cmux v0.1.5
 	github.com/stretchr/testify v1.8.1
 	github.com/uber/jaeger-client-go v2.28.0+incompatible
 	github.com/uber/jaeger-lib v2.2.0+incompatible
 	github.com/weaveworks/promrus v1.2.0
-	go.uber.org/atomic v1.5.1 // indirect
+	go.uber.org/atomic v1.5.1
 	golang.org/x/net v0.5.0
 	golang.org/x/tools v0.3.0
 	google.golang.org/grpc v1.53.0
--- a/middleware/grpc_instrumentation.go
+++ b/middleware/grpc_instrumentation.go
@@ -7,11 +7,13 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/atomic"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+
 	grpcUtils "github.com/weaveworks/common/grpc"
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -23,15 +23,16 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/prometheus/client_golang v1.14.0
+	github.com/prometheus/client_model v0.3.0
```
charleskorn, line 26: New dependency: used by tests as part of checking emitted metrics

go.mod
```diff
@@ -23,15 +23,16 @@ require (
...
 	github.com/sercand/kuberesolver/v4 v4.0.0
 	github.com/sirupsen/logrus v1.6.0
-	github.com/soheilhy/cmux v0.1.5 // indirect
+	github.com/soheilhy/cmux v0.1.5
 	github.com/stretchr/testify v1.8.1
 	github.com/uber/jaeger-client-go v2.28.0+incompatible
 	github.com/uber/jaeger-lib v2.2.0+incompatible
 	github.com/weaveworks/promrus v1.2.0
-	go.uber.org/atomic v1.5.1 // indirect
+	go.uber.org/atomic v1.5.1
```
charleskorn, line 35: Now depended on directly by `instrumentedClientStream`

go.mod
```diff
@@ -23,15 +23,16 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/prometheus/client_golang v1.14.0
+	github.com/prometheus/client_model v0.3.0
 	github.com/prometheus/exporter-toolkit v0.8.2
 	github.com/sercand/kuberesolver/v4 v4.0.0
 	github.com/sirupsen/logrus v1.6.0
-	github.com/soheilhy/cmux v0.1.5 // indirect
+	github.com/soheilhy/cmux v0.1.5
```
charleskorn, line 30: This is an unrelated change: I think this dependency changed from direct to indirect but `go.mod` wasn't updated at the same time.

charleskorn:
Will be moved to grafana/dskit once https://github.com/grafana/dskit/pull/342 is merged

charleskorn:
Closing in favour of https://github.com/grafana/dskit/pull/344

PR #299, opened 2023-07-25
Make gRPC logging optional via a custom interface

bboreham:
If an error returned to gRPC implements `Observe`, then that will get called and `GRPCServerLog` will do nothing.

This would replace #293.

### DIFF
--- a/middleware/grpc_logging.go
+++ b/middleware/grpc_logging.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
+	"errors"
 	"time"
 
 	"golang.org/x/net/context"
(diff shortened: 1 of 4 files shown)


middleware/grpc_logging.go
```diff
@@ -16,6 +17,11 @@ const (
 	errorKey = "err"
 )
 
+// If an error implements this interface, it will get called and GRPCServerLog will do nothing.
+type CustomLog interface {
+	LogOperation(ctx context.Context, _ logging.Interface, method string, duration time.Duration)
```
colega: Opinionated nit, I'd call this:
```suggestion
type Observer interface {
	Observe(ctx context.Context, _ logging.Interface, method string, duration time.Duration)
```

Why:
- `LogOperation` implies caller has to log an _operation_, but this is not something we require: they may or may not log it, we just want them to be aware of it, i.e. `Observe` it.
- `Observer` because that's idiomatic golang for single-method interfaces.

colega (approved):
LGTM, one small opinionated nit about naming.

middleware/grpc_logging.go
bboreham: OK I like `Observe` better, but does it convey that we no longer do the regular logging operation?

colega (approved):

